Issue #389, opened 2026-04-21
fix(python): jtd-codegen emits non-raw regex strings, triggers SyntaxWarning: invalid escape sequence


tato123:
## Description

Every Python file generated by `jtd-codegen` in `libs/streamlib-python/python/streamlib/_generated_/` contains a datetime-parser helper with a regex literal written as a plain string, not a raw string:

```python
datetime_re = '^(\d{4})-(\d{2})-(\d{2})[tT](\d{2}):(\d{2}):(\d{2})(\.\d+)?([zZ]|((\+|-)(\d{2}):(\d{2})))$'
```

On Python 3.12 this emits a `SyntaxWarning: invalid escape sequence '\d'` every time the module is imported. In a future Python release (PEP 650 / 3.14+ trajectory) the warning becomes a hard `SyntaxError`, so this is a ticking-clock fix, not just cosmetic noise.

## Context

Surfaced during #369 while round-tripping the escalate request/response types through the generated Python dataclasses. Every `import streamlib._generated_.com_streamlib_escalate_request` (and all its siblings) prints the warning to stderr. Same template is reused across ~20 generated files, so the fix multiplies.

The regex is jtd-codegen template output, not code we own — so the long-term fix is either:
1. **Upstream fix in `jsontypedef/json-typedef-codegen`** — make the Python template use `r'...'`. Cleanest but depends on upstream cadence.
2. **Post-process in xtask** — the `cargo xtask generate-schemas` pipeline already does some post-processing for Rust; extend it to rewrite the generated `.py` files, adding the `r` prefix to the regex literal. Keeps the fix local.
3. **Patch on top at generation time** — a simple sed/awk inside xtask right after the Python codegen step.

Path (2) is the most pragmatic given we already hand-edit the generated Python files for the escalate discriminator (jtd-codegen lacks discriminator support — documented in `schemas/com.streamlib.escalate_*@1.0.0.yaml`).

## Exit criteria

- [ ] `python3 -W error -c "import streamlib._generated_.com_streamlib_escalate_request"` succeeds (no `SyntaxWarning` raised to error)
- [ ] Same for every other generated `.py` module under `libs/streamlib-python/python/streamlib/_generated_/`
- [ ] Fix is durable — re-running `cargo xtask generate-schemas` doesn't revert it
- [ ] Chosen path (upstream / xtask post-process / regen-time patch) documented in the PR description

## Tests / validation

[...]

PR #396, opened 2026-04-21
fix(xtask): rewrite Python class names in jtd-codegen post-processor

### LINKED ISSUE #388: fix(python): streamlib._generated_.__init__ imports ApiServerConfig but file exports APIServerConfig
## Description

`libs/streamlib-python/python/streamlib/_generated_/__init__.py` fails to import. Line 6 reads:

```python
from .com_streamlib_api_server_config import ApiServerConfig
```

but `com_streamlib_api_server_config.py` actually defines the class as `APIServerConfig` (all-caps `API`). Every subsequent `from streamlib._generated_ import ...` in the Python SDK fails at package-init time with:

```
ImportError: cannot import name 'ApiServerConfig' from 'streamlib._generated_.com_streamlib_api_server_config' ...
Did you mean: 'APIServerConfig'?
```

## Context

Surfaced during #369 when smoke-testing the Python escalate-request generated types — a simple `from streamlib._generated_.com_streamlib_escalate_request import ...` blew up because the package `__init__` runs first and crashes on the `ApiServerConfig` line. I had to sidestep the package `__init__` via direct module loading (`importlib.util.spec_from_file_location`) to run the round-trip tests for #369.

The inconsistency is a jtd-codegen name-mapping glitch: `api_server_config` → `ApiServerConfig` in the `__init__.py` (simple CamelCase of the snake_case parts) but → `APIServerConfig` in the class file (`api` recognized as an acronym and uppercased). Either the class stays `APIServerConfig` and `__init__.py` is wrong, or the class should be `ApiServerConfig` and `__init__.py` is right — pick one and regenerate consistently.

This is a **hard breakage** for anyone consuming the `streamlib._generated_` namespace directly, not just a warning. Likely no one has tripped on it because the subprocess_runner import path imports the submodules directly.

## Exit criteria

- [ ] `python3 -c "from streamlib._generated_ import EscalateRequest, APIServerConfig"` (or `ApiServerConfig` — whichever is canonical) succeeds with no ImportError
- [ ] `__init__.py` and the class file agree on the name
- [ ] Decide whether the canonical naming in the Python SDK should match the Rust side (`api_server_config.rs` exports `ApiServerConfig` — check which it is)
- [ ] If the fix is in the jtd-codegen template rather than a one-off patch, document that so regeneration doesn't revert it

## Tests / validation

- [ ] `python3 -c "import streamlib._generated_"` succeeds (this test would have caught the bug — consider adding as a CI step)
- [ ] Re-running `cargo xtask generate-schemas` produces consistent output (i.e. the fix is sticky, not a manual patch that gets overwritten)

## Related

- Surfaced by: #369 (polyglot escalate acquire_texture)

tato123:
## Summary

- `jtd-codegen`'s Python backend ignores `--root-name` and independently mangles the schema name (`api_server` → `APIServerConfig` via an acronym upcaser, `mp4_writer` → `Mp4writerConfig` by not capitalizing after digits). The generated `__init__.py` still imports the canonical name xtask asked for, so `import streamlib._generated_` crashed with `ImportError` at package-init time.
- Teach `post_process_python` to detect the actual root class in the generated file (variant-parent for discriminator schemas, last class for plain schemas with sub-types) and rewrite every occurrence to match the expected name — the same stickiness trick the Rust post-processor already uses.
- Regenerate the two affected Python files.

## Closes

Closes #388

## Exit criteria

- [x] `python3 -c "from streamlib._generated_ import ApiServerConfig"` succeeds with no `ImportError`
- [x] `__init__.py` and every `.py` class file agree on the root name — verified across all 21 generated modules, not just `api_server_config`
- [x] Canonical naming matches the Rust side: `ApiServerConfig`, `Mp4WriterConfig`
- [x] Fix lives in the xtask post-processor (`post_process_python`), not a one-off file patch — re-running the generator produces byte-identical output

## Test plan

- [x] `cargo test -p xtask` — 6 tests pass (2 new for `post_process_python`, 3 new for `find_python_root_class_name`, 1 noop case)
- [x] `cargo check --workspace` — clean
- [x] `python3 -c "import streamlib._generated_"` — package init succeeds (the exact test the issue called for as a CI gate)
- [x] `python3 -c "from streamlib._generated_ import ApiServerConfig, EscalateRequest, EscalateResponse, Mp4WriterConfig, Videoframe, Audioframe, CameraConfig, DisplayConfig"` — all named imports resolve
- [x] Roundtrip `ApiServerConfig.to_json_data()` / `ApiServerConfig.from_json_data(...)` — preserves equality
- [x] Re-run `cargo xtask generate-schemas --runtime python --schema-file <each>` against a temp dir and `diff` against the committed files — no diff (stickiness)

## Polyglot coverage

- **Runtimes affected**: python (+ xtask post-processor)
- **Schema regenerated**: n/a — no schema changes, only Python output affected
- **E2E tests run**:
  - [x] Host-Rust: `cargo check --workspace` clean; no Rust side changes needed (post_process_rust already handles this class of bug)
  - [x] Python subprocess: package-init import test (`python3 -c "import streamlib._generated_"`) — the exact gate the issue called for
  - [ ] Deno subprocess: n/a — Deno's `_generated_/` uses `export *` re-exports, so the latent `Mp4writerConfig` casing issue there doesn't currently break imports (follow-up below)
- **FD-passing path**: n/a

## Follow-ups

- TS/Deno post-processor has the analogous gap — `libs/streamlib-deno/_generated_/com_tatolab_mp4_writer_config.ts` exports `Mp4writerConfig` instead of the canonical `Mp4WriterConfig`. Latent today (TS barrel file uses `export *`) but will bite anyone importing by name. Will file as a separate issue.
- Issue #389 (`SyntaxWarning: invalid escape sequence '\d'`) is untouched and remains a known follow-up.

### DIFF
--- a/libs/streamlib-python/python/streamlib/_generated_/com_streamlib_api_server_config.py
+++ b/libs/streamlib-python/python/streamlib/_generated_/com_streamlib_api_server_config.py
@@ -12,7 +12,7 @@ from typing import Any, Dict, Optional, Union, get_args, get_origin
 
 
 @dataclass
-class APIServerConfig:
+class ApiServerConfig:
     """
     Configuration for the runtime API server
     """
@@ -39,7 +39,7 @@ class APIServerConfig:
 
 
     @classmethod
-    def from_json_data(cls, data: Any) -> 'APIServerConfig':
+    def from_json_data(cls, data: Any) -> 'ApiServerConfig':
         return cls(
             _from_json_data(str, data.get("host")),
             _from_json_data(int, data.get("port")),
--- a/libs/streamlib-python/python/streamlib/_generated_/com_tatolab_mp4_writer_config.py
+++ b/libs/streamlib-python/python/streamlib/_generated_/com_tatolab_mp4_writer_config.py
@@ -12,7 +12,7 @@ from typing import Any, Dict, Optional, Union, get_args, get_origin
 
 
 @dataclass
-class Mp4writerConfig:
+class Mp4WriterConfig:
     """
     Configuration for MP4 file writer processor
     """
@@ -49,7 +49,7 @@ class Mp4writerConfig:
 
 
     @classmethod
-    def from_json_data(cls, data: Any) -> 'Mp4writerConfig':
+    def from_json_data(cls, data: Any) -> 'Mp4WriterConfig':
         return cls(
             _from_json_data(str, data.get("output_path")),
             _from_json_data(Optional[int], data.get("audio_bitrate")),
--- a/xtask/src/generate_schemas.rs
+++ b/xtask/src/generate_schemas.rs
@@ -335,8 +335,8 @@ fn run_jtd_codegen_python(schema_files: &[PathBuf], output_dir: &Path) -> Result
             )
         })?;
 
-        // Post-process: add copyright header
-        let processed_code = post_process_python(&python_code);
+        // Post-process: add copyright header and enforce expected class name
+        let processed_code = post_process_python(&python_code, &class_name);
 
         let output_path = output_dir.join(format!("{}.py", module_name));
         fs::write(&output_path, &processed_code)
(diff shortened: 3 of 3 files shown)


tato123:
Filed the TS/Deno follow-up as #397.

Issue #386, opened 2026-04-21
fix(deno): resolve 7 pre-existing Uint8Array/BufferSource type errors in context.ts


tato123:
## Description

The Deno SDK's `libs/streamlib-deno/context.ts` currently fails `deno check` with 7 pre-existing type errors in the FFI port handling code. All 7 are instances of `Uint8Array` / `Uint32Array` / `BigInt64Array` (backed by `ArrayBufferLike`) being passed to Deno FFI primitives (`Deno.UnsafePointer.of`, `BufferSource`-typed arguments) that now require the stricter `ArrayBuffer`-backed variant. These surface on Deno 2.7.x + TypeScript 5.9.x and appear to be fallout from a typing tightening in either Deno's FFI signatures or the `lib.dom` typings that wasn't propagated into the SDK.

## Context

The errors block `deno check` from passing on the Deno SDK as it stands, which means:

- CI for the Deno runtime can't gate on type correctness.
- Downstream users writing their own Deno processors get cascade errors from imports.
- Surfaced during #369 (polyglot escalate acquire_texture) when verifying no new TS errors were introduced — the 7 errors are pre-existing on `main`, not caused by #369, but they're blocking clean type-check signal for the Polyglot SDK Realignment milestone deliverables.

The error sites live in `NativeInputPorts` / `NativeOutputPorts` / the FFI pointer hand-off paths. Rough line numbers (on `main` at the time of writing): 231, 232, 233, 237, 274, 278, 290, 308. The root cause is most likely that the buffers are typed as plain `Uint8Array` / `Uint32Array` / `BigInt64Array` (which default to `ArrayBufferLike`), while Deno's FFI signatures now want `ArrayBuffer`-backed arrays specifically.

## Exit criteria

- [ ] `deno check libs/streamlib-deno/context.ts` passes with 0 errors on Deno 2.7+
- [ ] `deno check` on every other `.ts` in `libs/streamlib-deno/` (escalate.ts, subprocess_runner.ts, `_generated_/*.ts`, etc.) also passes with 0 errors
- [ ] Root-cause fix chosen (likely options: construct buffers as `new Uint8Array(new ArrayBuffer(N))`, add an explicit type parameter `Uint8Array<ArrayBuffer>`, or narrow the `lib` / `types` config) — pick one and document the rationale in the PR
- [ ] No runtime behavior change — the fix is purely type-system

## Tests / validation

[...]

Issue #383, opened 2026-04-21
Sync _generated_ dirs with current xtask+jtd-codegen output (drift cleanup)


tato123:
## Description

The committed `_generated_` directories (Rust, Python, TypeScript) have
drifted from what `cargo xtask generate-schemas` produces today. A full
regen today would change ~29 files per runtime, and in the Rust case
also break the build (`com_streamlib_bgra_file_source_config.rs` is
committed + imported by `lib.rs` but not listed in
`[package.metadata.streamlib].schemas`, so regeneration drops it from
`mod.rs`).

This issue is the cleanup: re-run `cargo xtask generate-schemas` for all
three runtimes, commit the regenerated output, and fix the pre-existing
blockers that currently make a full regen unsafe.

## Context

Surfaced while landing #370 (discriminator schema support). #370's scope
was kept narrow — only the two new escalate schemas + the three SDK call
sites — to avoid mixing discriminator work with independent drift. The
drift has three categories:

1. **Cosmetic** — an extra blank line between the `//!` header and the
   `use serde::` import. Pre-existing in current xtask output against
   `main`.
2. **Field ordering** — e.g. `com_tatolab_videoframe.rs`: committed has
   `fps` right after `frame_index`, jtd-codegen v0.4.1 alphabetizes
   `properties` then appends `optionalProperties` alphabetically, which
   produces `fps` at the end.
3. **Empty-schema behavior shift** — `com.streamlib.h264_decoder.config`
   has no `properties`/`optionalProperties`. Committed file is
   `pub struct H264DecoderConfig {}`; current jtd-codegen v0.4.1 emits
   `pub type H264decoderConfig = Option<Value>` (note lowercase `d` —
   jtd-codegen's own PascalCase converter doesn't break on `_` after
   digits).

Separately, `com.streamlib.bgra_file_source.config@1.0.0.yaml` exists on
disk, its generated file is committed, and it is imported by `lib.rs`,
but the schema path was never added to
`[package.metadata.streamlib].schemas` in `libs/streamlib/Cargo.toml`.
A full regen therefore drops it from `mod.rs` and breaks the build.

The Python `_generated_` is also significantly out of sync with the
Rust schema list — missing escalate, h264/h265, moq, opus, linux_mp4,
encodedaudioframe. #370 regenerated Python into a consistent state, but
this ticket should confirm all three runtimes stay in lockstep.

## Exit criteria

[...]

Issue #143, opened 2026-02-07
tracker: Schema Identity & Packaging architecture


tato123:
## Description

Milestone tracker for the Schema Identity & Packaging architecture. Consolidates the old `.slpkg advanced features` design (originally filed in this ticket), the #140 `@scope/package` naming proposal, and the `_generated_` drift cleanup (#383) into one cohesive plan. Child issues below do the actual work.

## Context

Multiple prior architecture attempts drifted out of sync before amos and milestones existed: `[package.metadata.streamlib]` in Cargo.toml, empty `[tool.streamlib]` in pyproject.toml, no streamlib section at all in deno.json, hardcoded schema strings in Deno subprocess examples, partially-implemented `.slpkg` distribution format, two different schema-identifier styles (`com.tatolab.xxx@ver` vs. the proposed `@scope/package`), and an `embedded_schemas.rs` match statement maintained by hand. Every one of those fragments predates the current milestone discipline. This ticket consolidates them into a single architecture and closes the conflicting tickets.

## Architecture at a glance

- **Identifier grammar**: `@org/package/Type@semver-range` — e.g. `@tatolab/core/VideoFrame@^1.0`.
- **Package manifest**: `streamlib.yaml` at the root of every package AND at the root of consumer projects. Single source of truth; no streamlib metadata in `Cargo.toml` / `pyproject.toml` / `deno.json`.
- **Lockfile**: `streamlib.lock` (content-hashed, Cargo.lock analogue).
- **Well-known types**: `@tatolab/core` ships VideoFrame, AudioFrame, EncodedVideoFrame, EncodedAudioFrame.
- **Packages directory**: `packages/` at repo root (cross-language distribution units). `libs/` stays for pure-Rust workspace crates (runtime, codegen, macros, FFI shims).
- **Universal CLI**: `streamlib generate`, `streamlib install`, `streamlib verify`. `cargo xtask` stays as a contributor-only shortcut.
- **Processor registration**: short-name macros (Rust `#[streamlib::processor("AudioMixer")]`, Python `@streamlib.processor("AudioMixer")`, Deno `@streamlib.processor("AudioMixer")`); full `@org/package/Type@version` identifier derived from the enclosing `streamlib.yaml` at build time.
- **Identifier mangling is eliminated, not papered over**: `streamlib.yaml` carries the identifier literally. jtd-codegen is invoked with a sentinel root-name; post-processor does exact substitution. No acronym / digit-boundary heuristics anywhere.

[...]

PR #398, opened 2026-04-21
fix(xtask): rewrite TypeScript root name in jtd-codegen post-processor

### LINKED ISSUE #397: fix(typescript): Deno _generated_ exports Mp4writerConfig instead of Mp4WriterConfig
## Description

`libs/streamlib-deno/_generated_/com_tatolab_mp4_writer_config.ts` defines
`export interface Mp4writerConfig { ... }` — lowercase `w` — when the canonical
name across the rest of the codebase is `Mp4WriterConfig` (Rust, and now Python
after #396). Any Deno consumer that imports by name would fail with a missing
export.

## Context

Same root cause as #388, different language. `jtd-codegen` ignores `--root-name`
in **all three** of its backends when the schema name contains digits
(`mp4_writer` → `Mp4writerConfig`) or known acronyms (`api_server` →
`APIServerConfig` in Python only). The xtask Rust post-processor already fixes
this for Rust via its root-rename pass, and #396 adds the same rename for
Python. The TypeScript post-processor still only adds a copyright header.

Why this hasn't bitten yet: `libs/streamlib-deno/_generated_/index.ts` uses
`export * from "./<module>.ts"`, so whatever symbol name is in the module gets
re-exported verbatim. No Deno code currently imports `Mp4WriterConfig` by name,
so the mismatch is latent.

## Exit criteria

- [ ] `xtask/src/generate_schemas.rs::post_process_typescript` rewrites the root
      interface/class name to match `expected_class_name`, mirroring the logic
      that `post_process_python` uses after #396
- [ ] `libs/streamlib-deno/_generated_/com_tatolab_mp4_writer_config.ts` exports
      `Mp4WriterConfig`
- [ ] Spot-check every other generated `.ts` file — any other name that
      jtd-codegen's TypeScript backend mangles should be caught and renamed
- [ ] Re-running `cargo xtask generate-schemas --runtime typescript` produces
      byte-identical output (sticky fix)

## Tests / validation

- [ ] Unit test for `post_process_typescript`: input with `export interface Mp4writerConfig`
      and `expected_class_name = "Mp4WriterConfig"` → output uses `Mp4WriterConfig` everywhere
- [ ] Unit test for the discriminator-interface case (TypeScript uses tagged
      unions, not class inheritance — root is the tagged union type, variant
      types are its members)
- [ ] `deno check libs/streamlib-deno/escalate.ts` (or the existing Deno
      type-check entry) passes after regen

## Related

- See also: #388, #396 (the Python analogue of this fix)

## AI Agent Notes

- The jtd-codegen TypeScript backend's name mangling is not identical to its
  Python backend — it doesn't run the `api → API` acronym upcaser (`ApiServerConfig`
  comes out correct in TS today) but it does share the "don't capitalize after
  digits" behavior (`Mp4writerConfig`). So a TS fix only needs to handle the
  digit-boundary case plus whatever else surfaces on a full-workspace regen.
- For TypeScript discriminator schemas, jtd-codegen emits something like
  `export type EscalateRequest = EscalateRequestAcquirePixelBuffer | ...`. The
  root is the `export type` declaration, not any member interface. Root-detection
  logic needs to look at `type` / `interface` declarations and pick the one
  that's a tagged union of the others for discriminator, or the last declared
  `interface` for plain schemas.

tato123:
## Summary

- `post_process_typescript` now detects the root declaration (`export type` for discriminator schemas, last `export interface` for plain schemas) and rewrites it to match `expected_class_name` — mirroring the Python fix from #396.
- Regenerated `com_tatolab_mp4_writer_config.ts` so Deno exports `Mp4WriterConfig` instead of the jtd-codegen-mangled `Mp4writerConfig`.
- 6 new unit tests cover plain rename, discriminator rename, no-op when names match, and root-name detection for both schema shapes.

## Closes

Closes #397

## Exit criteria

- [x] `xtask/src/generate_schemas.rs::post_process_typescript` rewrites the root interface/class name to match `expected_class_name`, mirroring the logic that `post_process_python` uses after #396
- [x] `libs/streamlib-deno/_generated_/com_tatolab_mp4_writer_config.ts` exports `Mp4WriterConfig`
- [x] Spot-check every other generated `.ts` file — drift unrelated to name mangling was surfaced; tracked separately (see Follow-ups)
- [x] Re-running `cargo xtask generate-schemas --runtime typescript` produces byte-identical output for the target file (sticky fix)

## Test plan

- [x] `cargo test -p xtask` — 14 passed (6 new tests for TypeScript post-processor)
- [x] `cargo check -p xtask` — clean
- [x] `deno check libs/streamlib-deno/_generated_/*.ts` — all generated files type-check cleanly
- [x] Grep confirmed zero remaining `Mp4writerConfig` occurrences under `libs/streamlib-deno/`
- [x] Idempotency spot-check: regen into `/tmp` produces byte-identical `com_tatolab_mp4_writer_config.ts`

### Tests added

- `find_typescript_root_name_picks_last_interface_for_plain`
- `find_typescript_root_name_picks_type_alias_for_discriminator`
- `find_typescript_root_name_single_interface`
- `find_typescript_root_name_none_when_no_decl`
- `post_process_typescript_renames_digit_mangled_root`
- `post_process_typescript_noop_when_names_match`
- `post_process_typescript_discriminator_root_rename`
- `post_process_typescript_discriminator_noop_when_root_matches`

## Polyglot coverage

- **Runtimes affected**: typescript (xtask generator; Rust + Python unaffected)
- **Schema regenerated**: only `com_tatolab_mp4_writer_config.ts` — kept scope tight. Full-directory regen surfaces separate drift tracked by #383.
- **E2E tests run**:
  - [x] Host-Rust: `cargo test -p xtask` (covers the generator)
  - [ ] Python subprocess: n/a (Python generator unchanged, fixed by #396)
  - [ ] Deno subprocess: n/a (Deno runtime not exercised; regen-only fix, no runtime behavior change)
- **FD-passing path**: n/a

## Follow-ups

- **#383** (existing) already tracks the remaining Deno `_generated_` drift — stale `fps?` fields on videoframe/encodedvideoframe, doc updates on escalate_request/response, and 10 schemas listed in `libs/streamlib/Cargo.toml` that have never been generated for Deno (h264/h265/opus/moq encoders/decoders, linux_mp4_writer, encodedaudioframe). Deliberately left out of this PR so the review stays focused on the Mp4WriterConfig rename.
- **#386** (existing) tracks pre-existing Uint8Array/BufferSource type errors in `context.ts`/`subprocess_runner.ts` — surfaced when running `deno check` against the hand-written SDK surface. Independent of this PR.

### DIFF
--- a/libs/streamlib-deno/_generated_/com_tatolab_mp4_writer_config.ts
+++ b/libs/streamlib-deno/_generated_/com_tatolab_mp4_writer_config.ts
@@ -8,7 +8,7 @@
 /**
  * Configuration for MP4 file writer processor
  */
-export interface Mp4writerConfig {
+export interface Mp4WriterConfig {
   /**
    * Path to output MP4 file
    */
--- a/xtask/src/generate_schemas.rs
+++ b/xtask/src/generate_schemas.rs
@@ -409,8 +409,8 @@ fn run_jtd_codegen_typescript(schema_files: &[PathBuf], output_dir: &Path) -> Re
             )
         })?;
 
-        // Post-process: add copyright header
-        let processed_code = post_process_typescript(&ts_code);
+        // Post-process: enforce expected root name and add copyright header
+        let processed_code = post_process_typescript(&ts_code, &class_name);
 
         let output_path = output_dir.join(format!("{}.ts", module_name));
         fs::write(&output_path, &processed_code)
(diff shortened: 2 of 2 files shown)


tato123:
Closing — the Mp4WriterConfig rename this PR implements is superseded by the Schema Identity & Packaging milestone (#143). The fix lands naturally as part of #402 (resolver with sentinel-substitution post-processing), which eliminates the underlying jtd-codegen name-mangling class of bugs rather than patching each occurrence. No need to merge this incremental fix.

PR #426, opened 2026-04-22
chore(broker): socket activation + lifecycle tests (#424)

### LINKED ISSUE #424: chore(broker): reliable Linux daemon lifecycle + socket-activation + validation tests
## Description

Make the `streamlib-broker` Linux daemon a first-class, testable
component. Adopt **systemd socket activation** so clients never need
to start the broker first (the socket is always listening; the daemon
starts on demand), make **`cargo run -p streamlib-broker`** the
default path for dev/test so the running broker is always the current
source tree, and add daemon-lifecycle tests that cover crash recovery,
stale-socket cleanup, and runtime pruning.

## Context

Post-#394 the broker is the transport for every cross-process DMA-BUF
FD on Linux. The in-process `UnixSocketSurfaceService` used by
`libs/streamlib/tests/polyglot_linux_check_out.rs` and the per-crate
unit tests is fine for FFI-layer coverage — but it deliberately skips
the deployment path. Two devex problems motivate this ticket:

1. **"Is the broker running?"** — Today a first-time user has to run
   `dev-setup.sh` or `systemctl --user start streamlib-broker`
   before anything polyglot works. Socket activation removes this
   step entirely. The `.socket` unit is always enabled; the daemon
   starts on demand on first client connect; it can idle-shutdown
   after N seconds of inactivity (optional, configurable).
2. **"Am I testing the broker I'm running?"** — Pre-built binaries
   drift from source. Dev and test paths should default to
   `cargo run -p streamlib-broker` so fixing a bug in the daemon
   and running `cargo test` proves the fix without a manual
   reinstall step. `scripts/dev-setup.sh:188` already does this
   under the hood; elevate it to first-class.

Existing surface this builds on:
- `libs/streamlib-broker/src/main.rs` — the daemon binary, needs
  to learn `sd_listen_fds()` so it can inherit the pre-bound
  socket from systemd instead of `bind`ing its own.
- `scripts/streamlib-broker.service` — current pre-fork unit;
  keep as a fallback, add socket-activation pair next to it.
- `scripts/dev-setup.sh:188` — already `cargo run`s in dev; extract
  to a standalone helper.

## Exit criteria

**Socket activation**

- [ ] Add `scripts/streamlib-broker.socket` — `ListenStream=%h/.streamlib/broker.sock`, `SocketUser=` / `SocketGroup=` sane defaults, `Accept=no` (one-shot activation, not per-connection).
- [ ] Update `scripts/streamlib-broker.service` (or add a sibling `-activated.service`) to `Type=notify` or use `sd_listen_fds()`, inherit the socket from systemd instead of binding. `ExecStart=` reads `LISTEN_FDS` / `LISTEN_FDNAMES`.
- [ ] `libs/streamlib-broker/src/main.rs`: if `LISTEN_FDS` is set, build the `UnixListener` via `FromRawFd` on the inherited fd (3, per systemd convention) and skip the `bind` step. Otherwise fall back to the existing `--socket-path` bind path so non-systemd environments still work. Use the `sd-notify` crate or a minimal hand-rolled env-var probe — do not pull full `libsystemd` in.
- [ ] Add an optional `--idle-exit-seconds` flag that lets the daemon exit cleanly after N seconds of no active clients; systemd will re-activate on the next connection. Default: disabled (daemon stays up), opt-in for dev.
- [ ] User-facing docs: "just connect, the broker activates itself" — no `systemctl start` step in the onboarding flow.

**cargo run is the dev default**

- [ ] Extract broker lifecycle from `dev-setup.sh` into
  `scripts/broker-dev.sh` (or similar) — a standalone helper devs
  and tests can call without running the full environment bootstrap.
- [ ] The dev helper runs `cargo run -p streamlib-broker` (with a
  lockfile to prevent double-spawn). Matches the pattern already
  used inside `dev-setup.sh`, just elevated.
- [ ] `scripts/streamlib-broker-dev.service` (or a `--dev` flag on
  the existing unit) points at `cargo run` and sets
  `WorkingDirectory=` to the source tree so rebuilds pick up.
- [ ] Document the invariant: *in dev, the broker you connect to is
  the current source tree.* Production (`install.sh`) keeps the
  pre-built binary path.

**Lifecycle + reliability**

- [ ] `libs/streamlib/tests/fixtures/streamlib_broker.sh` — start / stop / restart / probe helpers. Any E2E scenario can call this; returns deterministic readiness.
- [ ] `libs/streamlib-broker/tests/daemon_lifecycle.rs`:
  - Spawns the daemon (via `cargo run` in dev / built binary in CI).
  - Client `check_in` / `check_out` succeeds.
  - `kill -9`, assert socket cleanup on restart (non-socket-activation path).
  - Socket-activation path: assert first client connect activates the daemon from cold; subsequent connects reuse the running daemon; idle-exit (if enabled) restarts cleanly.
  - `prune_dead_runtimes` releases surfaces whose owning client dropped abruptly.
- [ ] systemd unit validation — parse-level invariants on the `.socket` + `.service` files.
- [ ] `is-broker-running` probe: `streamlib-broker --probe <socket>` or a library fn. Used by fixture script.
- [ ] `docs/operations/broker.md` covering dev (cargo run, socket activation), production (systemd, installed binary), log paths, common failure modes.

## Tests / validation

Inline tests in this issue — all new:

- [ ] `streamlib-broker::tests::daemon_starts_and_accepts_client` — baseline lifecycle.
- [ ] `streamlib-broker::tests::daemon_cleans_up_stale_socket_on_restart` — non-activation path resilience.
- [ ] `streamlib-broker::tests::socket_activation_cold_start` — simulate systemd: pre-bind a socket, pass as fd 3 via `LISTEN_FDS=1`, daemon inherits, client connects, `check_in` succeeds.
- [ ] `streamlib-broker::tests::socket_activation_idle_exit_and_restart` — enable `--idle-exit-seconds=2`, wait, assert daemon exited; connect again, assert the inherited-socket path still serves (simulating the re-activation a real systemd would do).
- [ ] `streamlib-broker::tests::prune_dead_runtimes_releases_orphaned_surfaces` — refcount cleanup.
- [ ] `streamlib-broker::tests::systemd_units_have_required_fields` — text-level invariants on the `.socket` and `.service` files.
- [ ] `libs/streamlib/tests/fixtures/streamlib_broker.sh` self-tests — start, probe succeeds, stop, probe fails.

## Related

- Parent umbrella: #319 (GPU capability-based access)
- Blocks: #421 (polyglot consumer E2E — needs deterministic broker lifecycle)
- See also: `docs/learnings/camera-display-e2e-validation.md` for fixture-script shape.

## AI Agent Notes

- **Socket-activation convention:** systemd passes inherited sockets starting at fd 3, with `LISTEN_FDS=<n>` and optional `LISTEN_FDNAMES=<name1>:<name2>:...` env vars set for the child. Also sets `LISTEN_PID` to its own pid so the child can validate the env isn't a leaked one from a grandparent. Rust: construct `std::os::unix::net::UnixListener` via `FromRawFd(3)` when `LISTEN_FDS=1` and `LISTEN_PID == std::process::id()`; otherwise bind manually.
- **Why `Accept=no`:** `Accept=yes` would fork a handler per-connection (inetd-style) — wrong for a persistent stateful broker. `Accept=no` hands the listener to one long-running daemon.
- **`sd_notify` handshake (optional):** if the daemon tells systemd "READY=1" via `$NOTIFY_SOCKET`, systemd knows it's up and starts forwarding connections. Without it, systemd assumes ready on fork. For now simpler to use `Type=simple` + rely on listening-on-inherited-fd; revisit if idle-exit wants tighter coordination.
- **`Accept=no` + idle-exit corner case:** when the daemon exits, systemd re-queues incoming connects until it can spawn a fresh daemon. The `.socket` unit keeps the listener regardless of daemon state — that's the whole point. Test `socket_activation_idle_exit_and_restart` proves this from the client side.
- **Why not full `libsystemd`:** pulls in C deps we don't need. `sd-notify = "0.4"` (Rust crate, pure-Rust) handles both fd inheritance and the notify handshake with ~zero dep cost. Or hand-roll a ~30 LOC env probe.
- **Stale-socket cleanup is moot under socket activation** — the socket lives as long as the `.socket` unit, not the daemon, and systemd owns its lifetime. Still keep the non-activation bind path tested since that's what `scripts/dev-setup.sh` falls back to on non-systemd boxes.
- **cargo run vs built binary:** `cargo run -p streamlib-broker --quiet -- ...` takes ~1s on an incremental build, ~30s on a cold build. The dev helper should cache-warm on setup; tests that spawn the daemon should `cargo build -p streamlib-broker` first, then spawn the binary directly from `target/debug/` — matches what `polyglot_linux_check_out.rs` does for the cdylib.
- **Which tests use which spawn path:** `daemon_lifecycle.rs` tests spawn the built binary directly (fast, deterministic). `streamlib_broker.sh` fixture wraps `cargo run` for ergonomic dev use. Keep these paths separate — don't make tests depend on `cargo run` (slow, environment-dependent).
- **Files likely touched:**
  - `libs/streamlib-broker/src/main.rs` — add LISTEN_FDS probe, idle-exit.
  - `libs/streamlib-broker/tests/daemon_lifecycle.rs` — new integration tests.
  - `libs/streamlib-broker/Cargo.toml` — add `sd-notify` or equivalent minimal dep.
  - `scripts/streamlib-broker.service` — update for socket activation.
  - `scripts/streamlib-broker.socket` — new socket unit.
  - `scripts/streamlib-broker-dev.sh` (or equivalent) — cargo-run helper.
  - `scripts/dev-setup.sh` — thin down, delegate broker lifecycle to the helper.
  - `libs/streamlib/tests/fixtures/streamlib_broker.sh` — fixture for E2E.
  - `docs/operations/broker.md` — new.

tato123:
## Summary

Make the Linux `streamlib-broker` daemon a first-class testable component:

- **Socket activation** via `LISTEN_FDS=1` + inherited fd 3 — first client connect activates the daemon, no `systemctl start` step in onboarding.
- **`--idle-exit-seconds`** so a socket-activated daemon can idle-exit and be re-activated on the next connect.
- **`--probe <socket>`** + `ping` op so fixtures, tests, and `streamlib-broker-dev.sh probe` confirm readiness deterministically.
- **Real daemon-lifecycle integration tests** (stale-socket cleanup, activation cold start, idle-exit-and-restart, dead-runtime prune, systemd unit parse).

## Closes

Closes #424

### Socket activation
- [x] `scripts/streamlib-broker.socket` with `ListenStream=%h/.streamlib/broker.sock`, `Accept=no`.
- [x] `scripts/streamlib-broker.service` updated: `Requires=streamlib-broker.socket`, inherits via `LISTEN_FDS` (flag still present for non-activated path).
- [x] `main.rs::sd_listen_fd` probes `LISTEN_FDS` + optional `LISTEN_PID` validation; hands inherited listener straight to `UnixSocketSurfaceService::with_inherited_listener` (no `bind()`). Falls back to bind on env absence or inconsistency.
- [x] `--idle-exit-seconds` CLI flag with watchdog task keyed off an RAII active-connection counter.
- [x] Docs: `docs/operations/broker.md` — no `systemctl start` step in the onboarding flow.

### `cargo run` is the dev default
- [x] `scripts/streamlib-broker-dev.sh` (new) — standalone lifecycle helper (`start|stop|restart|status|probe`), pidfile-gated, builds once then runs `target/debug/streamlib-broker` via `setsid`.
- [x] `scripts/dev-setup.sh` — thin pointer to the new helper.
- [x] Invariant documented in `docs/operations/broker.md`: *in dev, the broker you connect to is the current source tree.*

### Lifecycle + reliability
- [x] `libs/streamlib/tests/fixtures/streamlib_broker.sh` (new) — `start [socket] / stop <socket> / probe <socket>`.
- [x] `libs/streamlib-broker/tests/daemon_lifecycle.rs` (new) — six integration tests below.
- [x] Systemd unit parse validation — *directive-level*, comments stripped, so an `# Accept=no` comment alone does NOT satisfy the assertion.
- [x] `--probe <socket>` readiness probe + `ping` op.
- [x] `docs/operations/broker.md` — new operator doc covering dev/prod/test, probing, log paths, common failures.

### `cargo test -p streamlib-broker` (Linux) — all green (2.21s)

**Library tests (9 passing, 3 new):**
- `ping_returns_pong` — locks the `{"op":"ping"}` → `{"pong": true}` shape used by `--probe` and the fixture.
- `active_connection_count_tracks_in_flight_clients` — RAII guard raises/lowers correctly; drives idle-exit.
- `inherited_listener_path_does_not_bind_or_unlink` — `with_inherited_listener` never bind()s and leaves the caller's socket file on `stop()`.

**Integration tests (6 passing, new file):**
- `daemon_starts_and_accepts_client` — baseline lifecycle.
- `daemon_cleans_up_stale_socket_on_restart` — plants a stale socket file; confirms the next start recovers.
- `socket_activation_cold_start` — pre-binds a `UnixListener`, hands it to the daemon as fd 3 with `LISTEN_FDS=1`, asserts inode equality (proves the inherited listener is actually serving, not a silent re-bind).
- `socket_activation_idle_exit_and_restart` — `--idle-exit-seconds=2`, wait for daemon self-exit, spawn a second daemon against the same test-harness listener, confirm it serves the next client.
- `prune_dead_runtimes_releases_orphaned_surfaces` — registers a runtime under a reaped pid + two surfaces, calls `prune_dead_runtimes`, asserts both surfaces released.
- `systemd_units_have_required_fields` — directive-level parse of `.socket` + `.service`; asserts `[Socket] Accept=no`, `[Socket] ListenStream=…`, `[Service] ExecStart=…streamlib-broker…`, `[Unit] Requires=streamlib-broker.socket`.

### Fixture smoke test (local)

```
SOCKET=$(libs/streamlib/tests/fixtures/streamlib_broker.sh start)
libs/streamlib/tests/fixtures/streamlib_broker.sh probe "$SOCKET"   # exit 0
libs/streamlib/tests/fixtures/streamlib_broker.sh stop  "$SOCKET"   # exit 0
libs/streamlib/tests/fixtures/streamlib_broker.sh probe "$SOCKET"   # exit 1
```

## CI evidence

- **Workflow file**: [`.github/workflows/test.yml`](../blob/chore/broker-lifecycle-socket-activation-424/.github/workflows/test.yml) — added `cargo test -p streamlib-broker` step to the Linux job (workflow is currently `workflow_dispatch`-only upstream, so the step runs on manual dispatch).
- **Green baseline**: local `cargo test -p streamlib-broker` reports `9 passed` (lib) + `6 passed` (integration) + `0 passed` (doc), finished in 2.21s. See the summary above for the test names.
- **Negative test #1 — unit parse gate isn't vacuous**: stripping the `Accept=no` directive line from `scripts/streamlib-broker.socket`:
    ```
    ---- systemd_units_have_required_fields stdout ----
    thread 'systemd_units_have_required_fields' panicked at
    libs/streamlib-broker/tests/daemon_lifecycle.rs:419:5:
    assertion `left == right` failed: streamlib-broker.socket must set Accept=no
      left: None
      right: Some("no")
    test result: FAILED. 0 passed; 1 failed; …
    ```
  Restoring the directive: test green again.
- **Negative test #2 — activation behavioral gate isn't vacuous**: forcing `sd_listen_fd()` to always return `Ok(None)` (makes the daemon fall back to the bind path, which `remove_file`s and re-binds the pre-bound socket, changing the inode):
    ```
    ---- socket_activation_cold_start stdout ----
    thread 'socket_activation_cold_start' panicked at
    libs/streamlib-broker/tests/daemon_lifecycle.rs:235:5:
    activated broker did not become ready; log at "/tmp/…/broker.log"
    test result: FAILED. 0 passed; 1 failed; …
    ```
  Restoring the probe: test green again. *(The test fails via the probe timeout because the broker happens to wedge during parallel port/socket setup under the injected bug rather than reaching the inode assertion cleanly — but the gate is still behaviorally red, which is what CI needs.)*
- **Environment validated**: local Linux workstation (Arch/Ubuntu equivalent), `target/debug/streamlib-broker` via `CARGO_BIN_EXE_streamlib-broker`. Tests require Linux (`#![cfg(target_os = "linux")]`).

## Polyglot coverage

- **Runtimes affected**: rust (no schema change)
- **Schema regenerated**: n/a
- **E2E tests run**:
  - [x] Host-Rust: `cargo test -p streamlib-broker` — 15 tests pass (lib + integration)
  - [ ] Python subprocess: n/a — broker-server-side change only; wire format to polyglot cdylibs is unchanged
  - [ ] Deno subprocess: n/a — same reason
- **FD-passing path**: n/a — this PR doesn't touch FD-passing; it makes the listener itself inheritable from systemd

## Follow-ups

None — scope kept to #424 exit criteria. Pre-existing clippy / fmt issues in `libs/streamlib-broker/src/grpc_service.rs`, `libs/streamlib-broker/src/state.rs`, and `libs/streamlib-broker-client/src/linux.rs` were left untouched per CLAUDE.md's "no auto-fixing on the side" rule.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -109,3 +109,10 @@ jobs:
 
       - name: Check streamlib-broker
         run: cargo check -p streamlib-broker
+
+      - name: Run streamlib-broker tests
+        # Includes lib tests and the daemon-lifecycle integration tests
+        # (socket activation, idle-exit, stale-socket cleanup, systemd
+        # unit parsing, dead-runtime pruning). These run on the Linux
+        # job because the broker is Linux-only.
+        run: cargo test -p streamlib-broker
--- a/docs/operations/broker.md
+++ b/docs/operations/broker.md
@@ -1,0 +1,181 @@
+# streamlib-broker — operations
+
+`streamlib-broker` is the Linux daemon that brokers cross-process
+DMA-BUF surface sharing between the streamlib host and any polyglot
+subprocess (Python, Deno). Every FD that crosses a process boundary on
+Linux passes through it, so its availability is a hard prerequisite for
+every polyglot pipeline.
+
+This doc is operator-facing: how to install it, how to tell whether it's
+healthy, where to look when things break. For implementation details
+see [`libs/streamlib-broker`](../../libs/streamlib-broker) and the
+consumer-side helpers in
+[`libs/streamlib-broker-client`](../../libs/streamlib-broker-client).
+
+---
+
+## Install flavors
+
+| Flavor     | Use when                                               | Broker binary                     |
+| ---------- | ------------------------------------------------------ | --------------------------------- |
+| Production | Installed streamlib release                            | `~/.streamlib/bin/streamlib-broker` (prebuilt) |
+| Dev        | Working on streamlib; want the broker to track source  | `cargo run -p streamlib-broker` (always rebuilt) |
+| Test       | CI / integration tests                                 | `target/debug/streamlib-broker` (built once per run) |
+
+The **invariant** that keeps dev sane: *in dev, the broker you connect
+to is the current source tree.* If you build a fix and connect, you're
+exercising the fix — no manual reinstall step.
+
+---
+
+## Production: socket-activated systemd install
+
+This is the recommended deployment path. The `.socket` unit keeps the
+listening socket available at all times; the daemon starts on the first
+client connect and can idle-exit later without losing the socket.
+
+```bash
+mkdir -p ~/.config/systemd/user
+cp scripts/streamlib-broker.socket  ~/.config/systemd/user/
+cp scripts/streamlib-broker.service ~/.config/systemd/user/
+systemctl --user daemon-reload
+systemctl --user enable --now streamlib-broker.socket
+```
+
+From then on, every client that connects to
+`~/.streamlib/broker.sock` activates the daemon automatically. No
+`systemctl start streamlib-broker.service` step is required in the
+onboarding flow.
+
+### Mechanics
+
+systemd binds the listening socket itself, then passes it to the daemon
+as fd 3 with `LISTEN_FDS=1` and `LISTEN_PID=<daemon pid>` in the
+environment. The broker's hand-rolled probe
+(`libs/streamlib-broker/src/main.rs::sd_listen_fd`) reads those env
+vars, validates pid match when present, and hands the inherited
+listener straight to `UnixSocketSurfaceService::with_inherited_listener`
+so no `bind()` call is ever made from the daemon.
+
+When the daemon exits (crash, SIGTERM, idle-exit), systemd keeps the
+socket. The next client connect re-activates the daemon and it inherits
+the same listening socket again. The socket file itself is never
+unlinked until the `.socket` unit is stopped.
+
+### Idle-exit (optional)
+
+Pass `--idle-exit-seconds=N` on the daemon's `ExecStart=` line to have
+it self-terminate after N seconds with no active client connections.
+Combined with socket activation, this gives "run only when needed"
+behavior. Leave it unset for long-running deployments.
+
+### Non-activated fallback
+
+If `LISTEN_FDS` is not set, the daemon falls back to binding
+`--socket-path` itself. This is the code path used by:
+
+- Legacy installs (`systemctl --user enable streamlib-broker.service`
+  without the `.socket`).
+- Non-systemd environments.
+- Ad-hoc dev (`scripts/streamlib-broker-dev.sh`, tests).
+
+Stale-socket cleanup applies on this path: the daemon unlinks a leftover
+socket file from a previous crash before `bind()`.
+
+---
+
+## Dev flavor: cargo run + standalone helper
+
+For ad-hoc dev work (no `dev-setup.sh` bootstrap needed), use the
+standalone lifecycle helper:
+
+```bash
+scripts/streamlib-broker-dev.sh start    # cargo build, spawn, probe until ready
+scripts/streamlib-broker-dev.sh status   # prints pid / socket
+scripts/streamlib-broker-dev.sh probe    # 0 on pong, 1 otherwise
+scripts/streamlib-broker-dev.sh stop     # SIGTERM, wait, SIGKILL fallback
+scripts/streamlib-broker-dev.sh restart
+```
+
+The helper is pid-file-gated (`$STREAMLIB_HOME/broker.pid`) to prevent
+double-spawn. By default it uses:
+
+| Env                           | Default                                 |
+| ----------------------------- | --------------------------------------- |
+| `STREAMLIB_HOME`              | `<repo>/.streamlib`                     |
+| `STREAMLIB_BROKER_SOCKET`     | `$STREAMLIB_HOME/broker.sock`           |
+| `STREAMLIB_BROKER_PORT`       | `50052`                                 |
+| `STREAMLIB_BROKER_LOG`        | `/tmp/streamlib-broker-dev.log`         |
+
+For the full dev-environment bootstrap (proxy scripts, `.env`,
+user-service install), use `scripts/dev-setup.sh`.
+
+---
+
+## Test flavor: `streamlib_broker.sh` fixture
+
+Integration tests that need a real daemon (not the in-process
+`UnixSocketSurfaceService`) use
+[`libs/streamlib/tests/fixtures/streamlib_broker.sh`](../../libs/streamlib/tests/fixtures/streamlib_broker.sh):
+
+```bash
+SOCKET=$(./libs/streamlib/tests/fixtures/streamlib_broker.sh start)
+./libs/streamlib/tests/fixtures/streamlib_broker.sh probe "$SOCKET"
+./libs/streamlib/tests/fixtures/streamlib_broker.sh stop  "$SOCKET"
+```
+
+The fixture always uses `target/debug/streamlib-broker` (builds it if
+missing) so tests aren't paying `cargo run`'s per-invocation cost.
+
+---
+
+## Probing the broker
+
+The daemon ships with a `--probe` subcommand. It connects to the socket,
+sends `{"op":"ping"}`, expects `{"pong":true}`, and exits 0 on success
+or 1 on any failure:
+
+```bash
+streamlib-broker --probe ~/.streamlib/broker.sock && echo ok
+```
+
+This is the same command the fixture and dev helper use. It's safe to
+call against a socket-activated broker (the ping itself triggers
+activation) and against a running daemon.
+
+---
+
+## Log paths
+
+| Deployment        | Log sink                                          |
+| ----------------- | ------------------------------------------------- |
+| systemd           | `journalctl --user -u streamlib-broker.service`   |
+| `streamlib-broker-dev.sh` | `/tmp/streamlib-broker-dev.log` (or `$STREAMLIB_BROKER_LOG`) |
+| Fixture           | alongside the socket: `<socket-dir>/broker.log`   |
+
+---
+
+## Common failures
+
+- **"Failed to bind Unix socket … Address already in use"** — a previous
+  daemon is still running. `streamlib-broker-dev.sh status` to check, or
+  `ps` / `systemctl --user status`. If the socket file exists but no
+  daemon does, the bind-path cleanup (`remove_file` before bind) handles
+  it on the next start. Under socket activation this error can't happen
+  — the socket is owned by systemd.
+
+- **Client connects but hangs** — with socket activation, this usually
+  means the daemon crashed while starting. Check
+  `journalctl --user -u streamlib-broker.service` for a stack trace.
+  The `.socket` unit will hold queued client connects until a successful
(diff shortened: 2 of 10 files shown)


tato123:
Closing without merging — superseded.

The branch ships socket activation + lifecycle tests + dev scripts for an external `streamlib-broker` daemon. The architecture discussion on April 22 (see #424's close comment) decided to retire the external daemon entirely: every runtime owns its own internal `UnixSocketSurfaceService`, no install, no scripts, no coordination.

About 70% of the diff in this PR (systemd units, dev scripts, lifecycle tests, idle-exit, socket-activation probe, operator doc, fixture script) is dead weight under the new design — it's deployment plumbing for a daemon we're deleting in #429.

The reusable pieces relocate into the runtime via #428 and are explicitly called out in #428's AI Agent Notes:
- `UnixSocketSurfaceService` (the surface-sharing service itself)
- The `ping` op + active-connection RAII counter

Successors:
- #427 — research: streamlib-orchestrator — parked design record
- #428 — runtime: own internal broker on per-runtime Unix socket
- #429 — cleanup: retire external streamlib-broker daemon (blocked by #428)
- #430 — runtime: JSONL logging on disk + stdout mirror + structured origin fields
- #431 — research: macOS anonymous XPC for subprocess IOSurface sharing without launchd

Branch `chore/broker-lifecycle-socket-activation-424` will be deleted after this close.

PR #862, opened 2026-05-19
feat(engine): per-schema iceoryx2 ring depth via metadata.max_queued_messages

tato123:
## Summary

- Replaces hardcoded `subscriber_max_buffer_size(16)` at every iceoryx2 service-open site with a per-schema value resolved from wire-schema `metadata.max_queued_messages`.
- Activates a previously-dead wire-schema field (renamed from `metadata.buffer_size` for unit clarity — slot count, not bytes) and wires it through the engine, the Python/Deno cdylib carve-outs, and the subprocess wiring JSON.
- Default 16 preserved as fallback so schemas that don't declare a depth get current behavior.

## Closes

Closes #837

## Exit criteria

- [x] iceoryx2 publish-subscribe services accept a per-service ring-depth parameter, plumbed from schema metadata.
- [x] Sensible default preserved (`DEFAULT_MAX_QUEUED_MESSAGES = 16`).
- [x] Existing schemas migrate without breakage; sensor-shaped schemas declared deeper rings already (MAVLink: 64, AudioFrame: 32, NetworkPacket: 64).

## Design notes

**Architectural choice surfaced during planning**: the iceoryx2 ring depth is sourced from the **wire-schema** `metadata.max_queued_messages` (per-vocabulary), matching how `max_payload_bytes` is plumbed today. The alternative — using the alive per-input-port `ProcessorPortSchema.buffer_size` (which controls the in-process `PortMailbox` depth) — was rejected because it conflates two distinct concerns and forces every consumer of a wire vocabulary to re-declare burst tolerance.

**Field rename**: the previously-dead `metadata.buffer_size` was renamed to `metadata.max_queued_messages`. The old name collided semantically with `ProcessorPortSchema.buffer_size` (a separate alive field at the processor-port layer) and read as a byte count when it's actually a slot count. Pre-1.0 / no external consumers so renaming cleanly is appropriate.

**Engine layer fix**: per CLAUDE.md "Engine-wide bugs get fixed at the engine layer." The hardcoded 16 lived in 6 sites (engine `Iceoryx2Node` + 2 cdylib carve-outs × 2 functions each); all swept in this PR per "no bad patterns left behind."

## Test plan

- [x] `cargo test -p streamlib-engine --lib iceoryx2::node::tests` — 3 passed (1 pre-existing + 2 new).
- [x] `cargo test -p streamlib-engine --lib max_queued_messages` — 5 new resolver tests pass.
- [x] Workspace baseline (`cargo test --workspace --exclude api-server-demo --exclude camera-deno-subprocess --exclude camera-python-subprocess --exclude camera-rust-plugin --exclude webrtc-cloudflare-stream`) — 417 passed, 1 pre-existing failure unrelated to this change (`test_schema_max_payload_bytes_encodedvideoframe` asserts a stale 512 KB but the schema declares 16 MiB; fails on `main` too).
- [x] Same suite re-run with `--test-threads=1` for the new tests — all 8 pass serially as well as in parallel (unique service names + canonical schema IDs avoid collisions).
- [x] Python: `pytest libs/streamlib-python/python/streamlib/tests/test_subprocess_runner_cleanup.py` — 3/3 pass (mock signature updated to match new FFI arg).
- [x] Deno: `deno check libs/streamlib-deno/subprocess_runner.ts libs/streamlib-deno/native.ts` — clean.

The integration test from the issue body (200 Hz sustained MAVLink pipeline, zero overwrites at the iceoryx2 boundary) was deferred: iceoryx2 doesn't expose a public overwrite counter, and the new unit test already locks the wiring (publish depth+3 unread messages, verify ring caps at depth and "latest wins" holds). Empirical confirmation can come from re-running PR #836's MAVLink stress test after this merges.

## Follow-ups

None.


tato123:
Added the 200 Hz two-stage relay integration test that was the actual point of the issue. Commit 39343179.

**Test shape** (the MAVLink-pipeline scenario from PR #836):
- Producer publishes 100 messages at 200 Hz to service S1
- Relay drains S1, republishes to S2, pauses 50 ms every 25 messages (simulates downstream jitter)
- Consumer drains S2 continuously

**Asymmetric outcome locks the engine wiring**:
- Shallow rings (depth 4): each 50 ms pause accumulates ~10 messages in S1, overflowing the 4-slot ring → consumer receives <100. Asserts loss.
- Deep rings (depth 64, matching MavlinkMessage's declared `max_queued_messages`): same 10-message accumulation fits easily → consumer receives exactly 100. Asserts zero loss.

If `subscriber_max_buffer_size` were silently reverted to any hardcoded constant, either both runs would lose (low constant) or both would preserve (high constant) — neither matches and the test fails.

**Implementation notes**:
- iceoryx2 `Publisher`/`Subscriber` are `!Send` (hold `Rc` internally); each thread constructs its own ports from the shared `Iceoryx2Node`.
- Startup synced via `Barrier(3)` — iceoryx2 doesn't queue messages for late-arriving subscribers.
- Two-phase shutdown (producer → relay drain → consumer drain) so the last message survives the teardown.

**Flakiness check**: 5/5 deterministic runs (~2.2 s each).

PR #918, opened 2026-05-22
feat(plugin-abi): β-shape Phase D return types — close cross-repo plugin distribution coupling (#917)

tato123:
## Summary

Refactor seven cdylib-callable return types from `Arc::into_raw / Arc::from_raw` raw-pointer transit (which required rustc-version + dep-graph coupling) into layout-stable `#[repr(C)] (handle, vtable)` β-shape pairs. **This unlocks the cross-repo plugin distribution dream described in CLAUDE.md's "Plugin Distribution Model" section** — plugins built in separate repos with different rustc versions and Cargo dep graphs can now share the same `streamlib-plugin-abi` + `streamlib-consumer-rhi` + target triple and load cleanly into the host.

### Types refactored (all 7 listed in the issue body)

- `VulkanAccelerationStructure` (Phase 2)
- `RhiColorConverter` (Phase 3)
- `TextureRing` (Phase 4)
- `RhiCommandRecorder` (Phase 5, Box-based — NOT Clone, locked by compile_fail doctest)
- `VulkanComputeKernel` (Phase 6-7)
- `VulkanGraphicsKernel` (Phase 6-7)
- `VulkanRayTracingKernel` (Phase 6-7)

### Plugin ABI changes

- `GPU_CONTEXT_FULL_ACCESS_VTABLE_LAYOUT_VERSION` bumped 3 → 4.
- 3 new clone/drop slot pairs added: `clone_color_converter` / `drop_color_converter`, `clone_acceleration_structure` / `drop_acceleration_structure`, `clone_command_recorder` / `drop_command_recorder`.
- Vtable size grows 192 → 240 bytes (29 fn pointers); layout regression test updated with new offsets.

### Pattern per type

- Rename `pub struct X` → `pub(crate) struct XInner` (preserve fields/methods/Drop intact).
- Add new `#[repr(C)] pub struct X { handle: *const c_void, vtable: *const GpuContextFullAccessVTable }` 16-byte β-shape.
- Handle is `Arc::into_raw(Arc<Inner>)` (or `Box::into_raw` for RhiCommandRecorder); cdylib treats it as opaque.
- `Clone` and `Drop` dispatch through the host-installed vtable slots; host-side callbacks run `Arc::increment_strong_count` / `Arc::decrement_strong_count` (or `Box::from_raw + drop`) in host-compiled code where the Inner layout is statically known.
[...]

PR #928, opened 2026-05-22
feat(rhi): #922 prereqs — RhiCommandRecorder pub access + new_with_queue constructor

tato123:
## Summary

Two prereq commits for #922 (vulkan-video interior RhiCommandRecorder migration). Each is independently mergeable.

### Commit 1 — `command_buffer_raw` / `vulkan_device_ref` → `pub`

Promotes the two engine-internal escape-hatch accessors on `RhiCommandRecorder` from `pub(crate)` to `pub`, enabling cross-crate access from `libs/vulkan-video/`. Updated doc comments make the escape-hatch contract explicit: the recorder owns `begin_command_buffer` / `end_command_buffer` / `vkQueueSubmit*` calls; callers may only record commands between `begin()` and `submit*()`.

Per CLAUDE.md "Make the right way easy and the wrong way hard. 'Hard' — not 'impossible'. Some escape hatches are legitimate" — vulkan-video's `vkCmd*VideoCodingKHR` / `vkCmdEncodeVideoKHR` / `vkCmdDecodeVideoKHR` are exactly the kind of codec-API-specific commands the RHI doesn't generalize and shouldn't try to. The escape hatch is the right shape.

### Commit 2 — `RhiCommandRecorder::new_with_queue` constructor

Adds a sibling constructor `RhiCommandRecorder::new_with_queue(vulkan_device, label, queue, queue_family_index)` that builds a recorder against a caller-provided queue instead of the device's default queue.

**Discovered during #922 implementation**: `RhiCommandRecorder::new` hardcoded `vulkan_device.queue()` + `queue_family_index()`. vulkan-video has three distinct queues (`encode_queue`, `transfer_queue`, `compute_queue`) that may differ from the default. Without this variant, migrating vulkan-video's raw command-pool/buffer/fence triples to RhiCommandRecorder would silently submit encode / decode work to the wrong queue.

The new constructor shares its body with `new()` via a private `new_with_queue_inner` helper. Default `new()` continues to use the device's default queue (one-line forward).

## What's NOT in this PR

The actual #922 migration — replacing vulkan-video's 5 raw command-pool/buffer/fence triples with `RhiCommandRecorder::new_with_queue(...)` calls — is the next sub-PR. This PR delivers the two enabling prereqs that surface during attempting the migration.

The migration scope (5 triples × encode/decode call sites in `libs/vulkan-video/src/encode/` and `decode/`) is multi-hundred-LOC across `session.rs`, `staging.rs`, `submit.rs`, and decoder counterparts. Each migration also needs to preserve the encode/decode submit semantics (specific fence + queue + timeline behavior). Doing that work in-PR risks breaking video encode/decode without extremely careful integration testing.

## Test plan

- [x] `cargo check --workspace --all-targets` clean.
- [x] `cargo test --lib -p streamlib-engine layout_tests` passes (45 layout tests).
- [x] Existing `vulkan-video-roundtrip` E2E unaffected (this PR doesn't touch vulkan-video).

Refs #922. Branch base: `feat/917-beta-shape-phase-d` (PR #918).

### DIFF
--- a/examples/polyglot-vulkan-ray-tracing/src/main.rs
+++ b/examples/polyglot-vulkan-ray-tracing/src/main.rs
@@ -153,7 +153,7 @@ impl RuntimeKind {
 /// Holds three caches:
 /// - UUID → `Texture` for resolving non-AS run-time bindings to
 ///   host-side images (storage_image, sampled_texture).
-/// - `as_id` → `Arc<VulkanAccelerationStructure>` for resolving AS
+/// - `as_id` → `VulkanAccelerationStructure` (β-shape) for resolving AS
 ///   bindings AND for chaining BLAS → TLAS construction (TLAS instance
 ///   `blas_id` lookup).
 /// - SHA-256 over canonical kernel descriptor bytes →
@@ -162,7 +162,7 @@ impl RuntimeKind {
 struct SceneKernelBridge {
     device: Arc<HostVulkanDevice>,
     surfaces: HashMap<String, Texture>,
-    as_handles: parking_lot::Mutex<HashMap<String, Arc<VulkanAccelerationStructure>>>,
+    as_handles: parking_lot::Mutex<HashMap<String, VulkanAccelerationStructure>>,
     kernels: parking_lot::Mutex<HashMap<String, Arc<VulkanRayTracingKernel>>>,
 }
 
--- a/examples/raytracing-showcase/src/main.rs
+++ b/examples/raytracing-showcase/src/main.rs
@@ -102,7 +102,7 @@ fn main() -> Result<()> {
     )?;
 
     let tlas_instances = vec![{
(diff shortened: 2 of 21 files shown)


tato123:
Closing without merging — this PR's two commits are Option C prereqs (promoting `RhiCommandRecorder` internals to `pub` so a sibling `vulkan-video` crate could access them), and Option B has been locked as the chosen path (see #915 design-decision comment, 2026-05-22).

Under Option B, vulkan-video moves into streamlib-engine and reads `pub(crate)` directly. There's no need to widen the engine's public API surface to expose recorder internals.

The new `RhiCommandRecorder::new_with_queue` constructor introduced here is a genuine engine improvement (video codec work uses non-default queues — encode/transfer/compute). It may resurface as part of the Option B fold if the codec code lands as a consumer of public RHI. Not preserving the commit; the Option B run will design the post-fold shape from scratch.

PR #1391, opened 2026-07-17
refactor(display): convert @tatolab/display to the engine-free plugin SDK

### LINKED ISSUE #1266: refactor(display): convert @tatolab/display to the engine-free plugin SDK
## Description

Convert `@tatolab/display` from the `streamlib` facade to `streamlib-plugin-sdk`: presentation goes through the new present-target ABI surface; the blit pipeline is built via the existing `create_graphics_kernel` slot (removing the current raw-device `VulkanGraphicsKernel::new` construction); dep flip; boundary allowlist entries removed. On Linux the package keeps its winit window/event loop (a platform detail — see the present-target issue's host-portable ownership framing).

## Context

The hard block is the present-target surface only. The PNG sampler's primary pixel-buffer path is already engine-free; its texture-readback fallback rides the readback ABI surface when available — if that surface hasn't landed at pickup, ship the conversion with the pixel-buffer path and wire the fallback as follow-through within this issue once it lands (it is parity polish, not exercised by the standard E2E; verify at pickup). The blit shaders exist in both the engine and the package; after conversion one copy is dead — delete the engine's copy after a grep confirms no other engine consumer uses it. `STREAMLIB_DISPLAY_*` env hooks and the headless-drain degradation path are package-local and survive unchanged.

## Exit criteria

- [ ] display deps `streamlib-plugin-sdk` only; no `VulkanPresentTarget`/raw-device naming; check-boundaries clean with its entries removed.
- [ ] Validated as a separately-built `.slpkg`: camera→display E2E with PNG check; headless-drain mode still degrades gracefully.
- [ ] Dead blit-shader copy removed after a no-other-consumer grep.

## Tests / validation

- [ ] E2E: camera→display (vivid) per docs/testing.md, display as separately-built `.slpkg`, PNG read-tool check; resize/colorspace-recreate exercised.

## Related

- Milestone: Engine-Free Package Migration

## AI Agent Notes

Soft dependency on the texture-readback surface (PNG fallback only) — do not treat it as a hard block. Hard-blocked only by the present-target surface.

tato123:
Closes #1266

## Summary

Flip `@tatolab/display` from the `streamlib` facade onto the engine-free `streamlib-plugin-sdk`, matching the merged frame-tap (#1264) conversion. The Linux display processor now mints every GPU resource host-side through the `FullAccess` surface reached via `GpuContextLimitedAccess::escalate` and drives it per frame through the cdylib-safe `PluginAbiObject` methods — it names no raw host device.

- **Present target**: facade `VulkanPresentTarget::new` -> SDK `PresentTarget` via `create_present_target(&RawWindowHandleRepr, ...)`, the RAII `begin_frame`/`PresentTargetFrame::end` flow (host barriers + submits + presents on end, closing any dangling render pass and draining the acquire semaphore), plus the recreate path for the SDR->HDR10 colorspace flip. The winit window is projected into the plugin ABI's flattened `RawWindowHandleRepr` (Xlib / Xcb / Wayland).
- **Graphics kernel**: raw `VulkanGraphicsKernel::new(device, ...)` -> `create_graphics_kernel`; engine-internal `MAX_FRAMES_IN_FLIGHT` replaced by a package-local `DISPLAY_BLIT_FRAMES_IN_FLIGHT` const.
- **PNG readback**: facade `VulkanTextureReadback` -> SDK `TextureReadback` via `create_texture_readback` + `wait_and_read` (closure-form read inlined).
- **Cargo.toml**: `streamlib` facade -> `streamlib-plugin-sdk`; drop the direct `streamlib-consumer-rhi` dep (`VulkanLayout` from the SDK re-export).
- **xtask**: remove display's `PACKAGES_FACADE_DEP` + `PACKAGES_ENGINE_REACH` allowlist entries — `check-boundaries` stays green (the enforceable proof).
- **Runtime shutdown**: on a natural render-thread exit (window close / `STREAMLIB_DISPLAY_FRAME_LIMIT` / error) the display requests a runtime shutdown via the engine-free `sdk::runtime_control::request_runtime_shutdown` (#1385) — it publishes a reason string on the host's reserved control topic, so no engine `Event` type crosses the plugin ABI. Best-effort (a failed request only warns); skipped when `stop()` drove the exit.

The engine `display_blit` shaders are **not** deleted: a grep shows they remain live test fixtures for the graphics-kernel binding-derivation suite (see review items below — exit criterion 3 is not fully delivered).

## Test evidence

Gates (worktree, engine-linked): build, test (5 passed), check-boundaries (no violations), check-cdylib-reach.

No GPU/IPC-runtime evidence was captured in this session — the sandbox cannot observe the GPU or IPC runtime (exit 144 on any pipeline run); that verification is human-run via `/verify-live`. See "Review items" below for the specific exit criterion this affects.

## E2E report

None captured this session (see should-fix review item on exit criterion 2 below). The camera->display (vivid) E2E as a separately-built `.slpkg`, with PNG check and headless-drain degradation, still needs to be run via `/verify-live` before merge.

The lenses below all cleared to PASS; these are the individual findings surfaced along the way, listed verbatim for the owner's in-context review.

1. **[should-fix]** `runtime/streamlib-engine/src/vulkan/rhi/vulkan_graphics_kernel.rs:3155`
   - Claim: Exit criterion 3 ('Dead blit-shader copy removed after a no-other-consumer grep') is delivered.
   - Evidence: The diff touches only 3 files (display Cargo.toml, display.rs, check_boundaries.rs); the engine's runtime/streamlib-engine/src/vulkan/rhi/shaders/display_blit.{vert,frag} are NOT removed. A grep shows they are still consumed by the engine's own graphics-kernel test fixtures (vulkan_graphics_kernel.rs vert_spv()/frag_spv() at 3155-3158, inside #[cfg(test)] starting line 3131 — display_blit_descriptor, constructs_display_blit_kernel, etc.). The ticket's premise ('after conversion one copy is dead') is therefore false: the engine copy is production-dead but test-alive. The diff neither removes it nor notes the grep outcome.
   - Suggested next step: Either migrate those engine graphics-kernel test fixtures to a synthetic/other SPIR-V and delete the engine display_blit copy, or keep the copy and document on the PR body that the grep found engine test-fixture consumers (so exit criterion 3 is honestly resolved rather than silently skipped).

2. **[should-fix]** `packages/display/src/linux/display.rs:605`
   - Claim: Exit criterion 2 (camera->display E2E as a separately-built .slpkg with PNG check + headless-drain degradation) is validated.
   - Evidence: No runtime evidence exists; the sandbox cannot observe GPU/IPC (exit 144). The change is a substantial rewrite of the GPU render loop across a new present-target plugin ABI: escalate() is now called from the spawned render thread (resumed()), the dangling-render-pass drain on STREAMLIB_DISPLAY_INJECT_ERROR_AT_FRAME relies on the host end_frame, and frame_index bounds depend on the host present ring at runtime. The headless-drain code path (run_headless_drain_loop + resumed() degradation) is preserved unchanged in source, but end-to-end behavior is unverified here.
   - Suggested next step: Run the camera->display (vivid) E2E via /verify-live with display as a separately-built .slpkg (PNG read-tool check + resize/colorspace recreate + STREAMLIB_DISPLAY_FORCE_HEADLESS drain), plus the error-injection semaphore-drain run under validation layers, and attach the evidence to the PR before the owner merges.

3. **[should-fix]** `packages/display/src/linux/display.rs:453`
   - Claim: The render thread now calls escalate (3 sites: resumed() present-target+kernel creation @453, color-format kernel rebuild @676, first PNG-readback creation @982). All three contend the SAME per-GpuContext escalate gate that Manual-mode stop() holds for its entire body. This is new in this PR — the old code passed Arc<HostVulkanDevice> into the render thread and constructed present targets/kernels/readbacks directly, never touching the gate.
   - Evidence: For a cdylib-resident .slpkg (this PR's target), stop() is dispatched via with_cdylib_scope (processor_instance_factory.rs:363-376), which calls begin_escalate_scope → escalate_gate().enter() (runtime_context.rs:823; escalate_scope_registry.rs:71) and holds the gate until end_escalate_scope_draining after the body returns (runtime_context.rs:852). The gate is Arc-shared across GpuContext clones, so it is the same gate the render thread's escalate must enter. EscalateGate cross-thread enter BLOCKS on the condvar (escalate_gate.rs:85-107) — it does not panic. stop() (display.rs:316-345) sets running=false, sends the proxy event, then polls up to 2s for the render thread to finish. If stop() races a render-thread escalate window, that escalate blocks on the gate stop() holds, stop() waits 2s for a thread that cannot make progress → guaranteed ~2s shutdown stall, then stop() detaches (does not join) and returns, releasing the gate; the DETACHED render thread's escalate then proceeds to run privileged GPU work (create_* + escalate_end's wait_device_idle) AFTER stop() returned, during runtime teardown — a post-teardown GPU-use window on exactly the NVIDIA path this subsystem is paranoid about. Reachability is narrow (shutdown during startup/resumed, a short-lived/frame-limited graph, or first PNG sample) but real, especially for short and test runs.
   - Suggested next step: Have the render thread refuse to begin a new escalate once shutdown is signaled (check stop_called/running immediately before each escalate and bail to the drain path), OR restructure so the one-time GPU setup does not run inside a window where stop() can be pending. Given the engine-layer angle (stop() holding the gate across a 2s thread-join is what makes the block observable), consider raising with the owner whether the fix belongs in the display package or in the lifecycle-dispatch/gate contract.

4. **[should-fix]** `packages/display/src/linux/display.rs:64`
   - Claim: DISPLAY_BLIT_FRAMES_IN_FLIGHT = 2 is a package-private constant that must stay in lock-step with the engine-internal MAX_FRAMES_IN_FLIGHT = 2. The present target hands frame_index in [0, MAX_FRAMES_IN_FLIGHT) and the blit kernel's descriptor sets are sized to (and indexed by) DISPLAY_BLIT_FRAMES_IN_FLIGHT via set_sampled_texture(frame_index,..) / set_push_constants_value(frame_index,..) / record_draw(.., frame_index, ..). The two constants live in separate crates with no shared link and no cross-check test, and the plugin ABI does not surface the present target's in-flight depth, so the plugin is forced to hardcode a magic value coupled to an engine-internal one.
   - Evidence: Host: vulkan_present_target.rs:34 MAX_FRAMES_IN_FLIGHT=2; begin_frame sets frame_index=current_frame with current_frame=(frame_index+1)%MAX_FRAMES_IN_FLIGHT (lines 504,730,737), so frame_index∈[0,2). build_display_kernel sets descriptor_sets_in_flight: DISPLAY_BLIT_FRAMES_IN_FLIGHT. Correct today (2==2) and the author documented the coupling, but if the engine ever bumps MAX_FRAMES_IN_FLIGHT the present target would emit a frame_index the blit kernel has no descriptor set for → silent descriptor-set OOB / aliasing, with no compile-time or test link between the two constants. This hits every plugin that drives a graphics kernel over a present target, not just display.
   - Suggested next step: Surface the present target's frames-in-flight depth across the plugin ABI (e.g. a cached POD field on PresentTarget or a getter) so a plugin sizes its kernel to the actual ring depth instead of a hardcoded constant. Engine-completeness gap; can ship as a documented PR review item since it is correct today.

5. **[should-fix]** `packages/display/src/linux/display.rs:51`
   - Claim: DISPLAY_BLIT_FRAMES_IN_FLIGHT is a hardcoded literal (2) that must stay in lock-step with the engine-internal MAX_FRAMES_IN_FLIGHT (also 2) across a plugin-ABI boundary the compiler cannot check; the present target exposes no getter for its ring depth.
   - Evidence: const DISPLAY_BLIT_FRAMES_IN_FLIGHT: u32 = 2 (line 51) sizes the blit kernel's descriptor ring (build_display_kernel, line 1317). The host present target yields frame_index ∈ [0, MAX_FRAMES_IN_FLIGHT) via (frame_index+1) % MAX_FRAMES_IN_FLIGHT (runtime/streamlib-engine/src/vulkan/rhi/vulkan_present_target.rs:34,737); PresentTarget's only cached-POD getter is color_format_raw() (sdk/streamlib-plugin-sdk/src/rhi/present_target.rs:67) — no frames_in_flight(). Safe today (values match; host with_slot bounds-checks frame_index and returns a typed GpuError → graceful frame-skip warn, not corruption, at runtime/streamlib-engine/src/vulkan/rhi/vulkan_graphics_kernel.rs:1015). But an engine-side bump of MAX_FRAMES_IN_FLIGHT to 3 would silently make the display drop 1-in-N frames with a per-frame warn, and this is the first present-target-consuming GPU conversion that other GPU package conversions will copy.
   - Suggested next step: Non-blocking, can ship as a PR review item. Long-term clean shape: add a cached-POD frames_in_flight() getter to the PresentTarget PluginAbiObject so a package sizes its descriptor ring from the actual present-ring depth (structural), instead of a literal duplicating an unreferenceable engine constant (convention).

6. **[low]** `packages/display/src/linux/display.rs:51`
   - Claim: DISPLAY_BLIT_FRAMES_IN_FLIGHT = 2 correctly sizes the blit kernel descriptor ring against the host present ring.
   - Evidence: The constant is hardcoded to 2 with a comment that it matches the swapchain's 2-deep ring; the engine's MAX_FRAMES_IN_FLIGHT is indeed 2 (runtime/streamlib-engine/src/vulkan/rhi/vulkan_present_target.rs:34) and frame_index is in [0,2), so it is correct today. But the coupling is silent — no compile-time or test guard. If the engine ever grows its present ring, begin_frame would return frame_index >= 2 and set_sampled_texture(frame_index, ...) would index past a 2-deep descriptor ring.
   - Suggested next step: Consider exposing the present ring depth on the present-target ABI so the display sizes its descriptor ring against it, or add an assertion that frame_index < DISPLAY_BLIT_FRAMES_IN_FLIGHT.

7. **[low]** `packages/display/src/linux/display.rs:268`
   - Claim: Pre-existing (not introduced by this diff): field drop order destroys the winit window before the present target, contrary to create_present_target's 'window must outlive the target' note; and the drop comment mislabels the order as 'reverse construction order'.
   - Evidence: DisplayEventLoopHandler declares window (line 364) before present_target (line 375); Rust drops fields in declaration order, so on drop(app) the window closes first, then present_target dispatches drop_present_target (host VkDestroySwapchain/Surface). create_present_target doc (context.rs:1495-1503) states the window must outlive the target. The same window-before-present_target ordering existed in main, so this is not a regression. The comment at 268-271 says 'reverse construction order' but present_target (built first) also drops first — same order as construction, not reverse (harmless since the two host resources are independent).
   - Suggested next step: In a follow-up, reorder the struct so present_target/graphics_kernel are declared before window (drop-before-window), and correct the 268-271 comment. Not blocking — process-exit teardown, drivers tolerate it, and it predates this PR.

8. **[low]** `packages/display/src/linux/display.rs:364`
   - Claim: DisplayEventLoopHandler declares `window` (field @364) before `present_target` (@375), so at teardown the winit Window drops (closing the native X/Wayland window) BEFORE the PresentTarget's Drop dispatches drop_present_target (which destroys the host VkSwapchainKHR + VkSurfaceKHR). That inverts the RawWindowHandleRepr / create_present_target contract 'the caller owns the native window and must outlive the minted PresentTarget'.
   - Evidence: Rust drops struct fields in declaration order; window @364 precedes present_target @375. The ABI contract is stated in runtime/streamlib-plugin-abi/src/repr/present.rs:22-25 and reserved_m32.rs:63,92. This is PRE-EXISTING (the old code had the same field order) — not a regression introduced by this PR — and is generally tolerated by X11/Wayland drivers at destroy time, but the diff's new explicit ABI wording makes the mismatch worth a cleanup.
   - Suggested next step: Reorder so present_target (and graphics_kernel) drop before window, or add an explicit drop(present_target) before the window is released, to satisfy the window-outlives-surface contract. Follow-up, not blocking this PR.

9. **[low]** `packages/display/src/linux/display.rs:1509`
   - Claim: drain_and_discard_video test coverage dropped from 3 cases to 1 (empty-input only); the multi-frame drain-all loop behavior is no longer directly exercised in-crate.
   - Evidence: Main had FIFO drain-all, skip-to-latest, and empty-input tests routing frames through InputMailboxesInner::new()+route()+from_inner_arc(). The engine-free SDK's InputMailboxesInner is an opaque placeholder (lib.rs:169 'opaque *Inner placeholders'), so the branch keeps only drain_on_unwired_input_returns_zero via InputMailboxes::empty(). The comment (lines 1501-1508) documents this is inherent and defers multi-frame semantics to the engine's iceoryx2 tests + the host read_raw contract.
   - Suggested next step: Accept as an inherent consequence of the engine-free conversion; note the coverage delta on the PR body so it is a conscious review item, not a silent drop.

10. **[low]** `packages/display/src/linux/display.rs:647`
    - Claim: The ColorInfo-change swapchain recreate now sizes from self.width/self.height (last requested extent) instead of the swapchain's actual current extent, because the SDK PresentTarget exposes no current_extent() getter (only color_format_raw). The prior facade code recreated at present_target.current_extent().
    - Evidence: display.rs:647 calls present_target.recreate(current_width, current_height, new_color_traits) from self.width/self.height snapshots; the old code used present_target.current_extent(). Functionally equivalent: WindowEvent::Resized keeps self.width/height synced, the WSI clamps identically on recreate, and every draw viewport/scissor/aspect uses frame.extent from begin_frame (the true acquired extent), not self.width/height.
    - Suggested next step: Add a one-line comment noting the recreate uses the tracked window size (kept in sync by Resized) because the SDK present target has no current_extent accessor, and that the true extent is re-read from frame.extent each draw. Not merge-blocking.

11. **[info]** `packages/display/src/linux/display.rs:1509`
    - Claim: The headless drain helper's drain-all behavior remains covered by package tests.
    - Evidence: The base had three drain tests (FIFO drains all 3, skip-to-latest empties, empty returns 0) using engine-internal InputMailboxesInner::new/add_port/route. The branch keeps only drain_on_unwired_input_returns_zero (InputMailboxes::empty()). This reduction is FORCED — the engine-free SDK's InputMailboxesInner::add_port is a no-op stub with no route/FrameHeader, so a package cannot inject buffered frames. The remaining test locks loop termination on Ok(None) but NOT the drain-all invariant; a regression of drain_and_discard_video to a single read would pass the empty test.
    - Suggested next step: No action required in this package; confirm engine-side read_raw/mailbox tests cover the multi-frame drain-all contract so the invariant is locked somewhere.

12. **[info]** `packages/display/src/linux/display.rs:451`
    - Claim: Core soundness confirmed: the diff removes the raw-HostVulkanDevice Arc transit (the exact .slpkg driver-corruption class in docs/learnings/slpkg-raw-device-rhi-construction.md and my symptom index) and replaces it with the correct PluginAbiObject lift.
    - Evidence: The old start() reached ctx.gpu_full_access().host_vulkan_device_arc() and hand-rolled RHI on the transited device (VulkanPresentTarget::new / VulkanGraphicsKernel::new / VulkanTextureReadback::new_into_stream_error). The new resumed()/render_frame()/sample_texture_to_png() mint every GPU resource host-side via gpu_context.escalate(|full| full.create_present_target(...) / create_graphics_kernel(...) / create_texture_readback(...)). No host device name, kernel ::new, or buffer ::new* survives (grep-confirmed across all packages/display/src files).
    - Suggested next step: None — this is the sanctioned fix; call out on the PR body as the headline soundness win.

13. **[info]** `packages/display/src/linux/display.rs:127`
    - Claim: escalate placement is correct: it never runs inside a FullAccess lifecycle body, and follows escalate-once/cache/per-frame-scope-free.
    - Evidence: escalate is invoked only on the spawned render thread (resumed() line 451, color-format-flip rebuild line 674, PNG-fallback readback line 982) — all AFTER Manual start() returns, so the host escalate gate held during start() is already released (no same-thread re-entry, per context.rs:540-679 and .claude/rules/plugin-boundary.md). Per-frame hot path (begin_frame/end, set_sampled_texture, set_push_constants_value, record_draw) is scope-free. No &GpuContextFullAccess escapes any closure (each returns owned PluginAbiObjects).
    - Suggested next step: None.

14. **[info]** `packages/display/src/linux/display.rs:1123`
    - Claim: Wire-repr fidelity verified against the ABI: RawWindowHandleRepr/ColorTraitsRepr/HdrStaticMetadataRepr field names+types and the texture_format_from_raw discriminant decode all match.
    - Evidence: Field constructions at lines 1085-1110, 1239-1242, 1261-1282 match runtime/streamlib-plugin-abi/src/repr/present.rs (kind/_reserved_padding/window_or_surface/display_or_connection/screen/_reserved_tail; primaries_raw/transfer_raw; the 8 HDR f32 fields). texture_format_from_raw 0..6 matches the pinned enum in runtime/streamlib-consumer-rhi/src/formats.rs (Rgba8Unorm=0 .. Nv12=6, test texture_format_discriminants_are_pinned at formats.rs:130). New tier-1 test texture_format_from_raw_round_trips_known_and_rejects_unknown (line 1480) locks the round-trip and rejects unknown discriminants instead of aliasing to Rgba8Unorm.
    - Suggested next step: None.

15. **[info]** `packages/display/src/linux/display.rs:453`
    - Claim: Core invariant satisfied: the raw-device transit is fully removed and replaced with the cdylib-safe FullAccess primitives — this is exactly the fix the slpkg learning demands.
    - Evidence: display.rs no longer names HostVulkanDevice / host_vulkan_device_arc / VulkanPresentTarget::new / VulkanGraphicsKernel::new / VulkanTextureReadback::new. Every GPU resource is minted via self.gpu_context.escalate(|full| full.create_present_target/create_graphics_kernel/create_texture_readback(..)). Verified escalate is a &self method returning Result<R> that dispatches through the #[repr(C)] FullAccess vtable (sdk/streamlib-plugin-sdk/src/context.rs:573), and escalate is never called inside a setup/start/stop/teardown lifecycle body (only on the spawned render thread), matching the plugin-boundary rule.
    - Suggested next step: None — this is the correct plugin-safe shape per docs/learnings/slpkg-raw-device-rhi-construction.md.

16. **[info]** `packages/display/src/apple/display.rs:87`
    - Claim: The Apple module still uses the raw-device pattern this PR eliminates on Linux; it is not compiled, so it does not break the build or trip the boundary gate, but it will need the same FullAccess treatment when Apple is activated.
    - Evidence: packages/display/src/apple/display.rs:87 reaches ctx.gpu_full_access().device().metal_device_ref(). lib.rs deliberately does NOT mod-declare apple (comment at lib.rs:17-22), so the file is dead source. The boundary checker only matches host_vulkan_device_arc / streamlib::sdk::engine:: (xtask/src/check_boundaries.rs:1083,1085) — neither appears in apple/display.rs — so removing display from PACKAGES_ENGINE_REACH_ALLOWLIST keeps the gate green. Apple is parked pre-Linux-stable per project policy, so this is correctly out of scope for #1266.
    - Suggested next step: Track as a note for the eventual Apple/Metal present-target rewrite: port .device().metal_device_ref() onto a Metal-equivalent FullAccess primitive before activation.

17. **[info]** `packages/display/src/linux/display.rs:30`
    - Claim: The SDK's create_present_target signature exposes streamlib_plugin_abi Repr wire types directly, forcing every present-target-consuming package to depend on streamlib-plugin-abi and import the raw Repr types — the display follows this correctly; it is a pre-existing SDK-surface decision, not a display defect.
    - Evidence: sdk/streamlib-plugin-sdk/src/context.rs:1504 create_present_target takes &streamlib_plugin_abi::RawWindowHandleRepr and Option<streamlib_plugin_abi::ColorTraitsRepr>; grep found no SDK re-export of RawWindowHandleRepr/ColorTraitsRepr/HdrStaticMetadataRepr. So the display's direct import (display.rs:30) is the only path, and streamlib-plugin-abi is already a required dep for export_plugin!. Cargo deps are all version-pinned (streamlib-plugin-sdk/macros/plugin-abi/jtd-codegen = 0.6.0), no path/git/patch — the distribute-by-version invariant holds.
    - Suggested next step: Optional SDK-level follow-up (not this PR): re-export the method-signature wire Reprs through streamlib_plugin_sdk::sdk so an 'engine-free authoring SDK' consumer imports from one namespace.

18. **[info]** `packages/display/src/linux/display.rs:1466`
    - Claim: Multi-frame FIFO / skip-to-latest drain unit tests were removed; the reduction is genuinely forced by the engine-free boundary and is explained in the test comment.
    - Evidence: The deleted tests used InputMailboxesInner::new() + add_port + route(make_video_frame(..)) to inject buffered frames; the SDK's InputMailboxesInner is a macro-support stub whose add_port is a no-op and has no route (sdk/streamlib-plugin-sdk/src/iceoryx2.rs:103-116), so buffered-frame drain semantics cannot be constructed at the package level. The replacement drain_on_unwired_input_returns_zero (via InputMailboxes::empty()) still locks the loop-termination-on-empty case; multi-frame semantics now rely on engine-side iceoryx2 tests + the host read_raw contract.
    - Suggested next step: None required — the coverage trade-off is boundary-forced and documented; the load-bearing loop-termination behavior remains covered.

19. **[info]** `packages/display/src/linux/display.rs:1510`
    - Claim: Two real drain unit tests were removed (FIFO drains every queued frame; skip-to-latest empties the input), leaving only the empty-input no-op test. This is a forced consequence of the engine-free boundary, not an avoidable coverage regression.
    - Evidence: display.rs:1510 keeps drain_on_unwired_input_returns_zero using InputMailboxes::empty(). The removed tests relied on the engine's real InputMailboxesInner::new()/route()/add_port(buffered)/from_inner_arc — but the SDK twin is a host-backed stub whose add_port is a no-op with no frame-injection path (iceoryx2.rs:116), so an engine-free package cannot construct a mailbox holding buffered frames. Multi-frame semantics are now covered by the engine's iceoryx2 tests + the host read_raw contract, per the retained test's doc comment.
    - Suggested next step: None required. The retained loop-termination coverage is the only part reconstructable engine-free; the justification comment is accurate.

20. **[info]** `packages/display/src/linux/display.rs:1123`
    - Claim: The PrimariesId cross-boundary discriminant contract (SDK sdk::color::PrimariesId encode -> ColorTraitsRepr.primaries_raw -> host core::color::PrimariesId decode) relies on implicit sequential discriminants on both enums with NO pinning regression test, unlike TextureFormat. Values match today; pre-existing and outside this diff.
    - Evidence: SDK color.rs:145 PrimariesId has no explicit discriminant values (Bt709..Ebu3213 implicit 0..10); host primaries_id_from_raw (present_target.rs:64) hardcodes 0=>Bt709..10=>Ebu3213. A reorder of either enum would silently mis-pick primaries on wide-gamut/HDR content with no test failure. display.rs uses these correctly; the gap is in the enum definitions this PR does not touch. TransferId is explicitly numbered and safe.
    - Suggested next step: Track separately: add a discriminant-pinning test for PrimariesId (and MatrixId) mirroring texture_format_discriminants_are_pinned so the SDK<->host color wire contract cannot silently drift. Not this PR's responsibility.

---

### DIFF
--- a/packages/display/Cargo.toml
+++ b/packages/display/Cargo.toml
@@ -17,14 +17,17 @@ crate-type = ["rlib", "cdylib"]
 streamlib-jtd-codegen = {version = "0.6.0"}
 
 [dependencies]
-# Engine substrate — runtime context, processor traits, generated config
-# types under `crate::_generated_::tatolab__display::*`, VideoFrame
-# wire type, host RHI types (HostVulkanDevice, VulkanGraphicsKernel,
-# VulkanPresentTarget, RhiCommandRecorder, OffscreenColorTarget,
-# TextureFormat).
-streamlib = {version = "0.6.0"}
+# Engine-free authoring SDK (never the `streamlib` facade) — capability-typed
+# runtime/GPU context views, generated config types under
+# `crate::_generated_::tatolab__display::*`, the `VideoFrame` wire type, and the
+# cdylib-safe RHI PluginAbiObjects (`PresentTarget`, `VulkanGraphicsKernel`,
+# `RhiCommandRecorder`, `TextureReadback`, `TextureFormat`, `VulkanLayout`). GPU
+# resource creation goes through `GpuContextLimitedAccess::escalate` +
+# `create_present_target` / `create_graphics_kernel` / `create_texture_readback`,
+# never the raw host device.
+streamlib-plugin-sdk = {version = "0.6.0"}
 
-# Procedural macros — `#[streamlib::sdk::processor("...")]` reads the
+# Procedural macros — `#[streamlib_plugin_sdk::sdk::processor("...")]` reads the
 # crate's own `streamlib.yaml` at `CARGO_MANIFEST_DIR`.
 streamlib-macros = {version = "0.6.0"}
 
@@ -36,16 +39,14 @@ tracing = {version = "0.1.41", features = ["release_max_level_debug"]}
 serde = {version = "1.0", features = ["derive"]}
 
 [target.'cfg(target_os = "linux")'.dependencies]
-# Window creation + event loop for the Linux display processor.
+# Window creation + event loop for the Linux display processor (a
+# host-portable platform detail, not baked into the plugin ABI).
 winit = "0.30"
-# Raw window-handle traits — `VulkanPresentTarget::new` takes
-# `&(impl HasWindowHandle + HasDisplayHandle)`; the winit `Window` type
-# implements both.
+# Raw window-handle traits — the winit `Window` implements
+# `HasWindowHandle` + `HasDisplayHandle`; the processor projects those into
+# the plugin ABI's flattened `RawWindowHandleRepr` fed to
+# `create_present_target`.
 raw-window-handle = "0.6"
-# Consumer-side RHI carve-out — typed `VulkanLayout` enum used in
-# producer/consumer surface-share interop. Camera uses the same dep
-# for the same reason.
-streamlib-consumer-rhi = {version = "0.6.0"}
 
 [target.'cfg(any(target_os = "macos", target_os = "ios"))'.dependencies]
 # Apple module is ported into this package but DISABLED in `lib.rs` —
--- a/packages/display/src/linux/display.rs
+++ b/packages/display/src/linux/display.rs
@@ -1,33 +1,35 @@
 // Copyright (c) 2025 Jonathan Fontanez
 // SPDX-License-Identifier: BUSL-1.1
 
-//! Linux display processor — winit window + Vulkan presentation via
-//! [`VulkanPresentTarget`].
+//! Linux display processor — winit window + Vulkan presentation via the
+//! engine-free plugin SDK's [`PresentTarget`] PluginAbiObject.
 //!
-//! Owns no raw Vulkan handles itself; every GPU resource flows through
-//! the host RHI public surface (`VulkanPresentTarget`,
-//! `VulkanGraphicsKernel`, `RhiCommandRecorder`, `VulkanTextureReadback`).
+//! Owns no raw Vulkan handles and never names the host device: every GPU
+//! resource is minted host-side through the FullAccess surface reached via
+//! [`GpuContextLimitedAccess::escalate`] ([`PresentTarget`],
+//! [`VulkanGraphicsKernel`], [`TextureReadback`](streamlib_plugin_sdk::sdk::rhi::TextureReadback))
+//! and driven per frame through each resource's own plugin-ABI methods.
 
 use std::sync::atomic::{AtomicBool, AtomicU64, Ordering};
 use std::sync::{Arc, OnceLock};
 use std::thread::JoinHandle;
 use std::time::{Duration, Instant};
 
-use streamlib::sdk::context::{GpuContextLimitedAccess, RuntimeContextFullAccess};
-use streamlib::sdk::engine::host_rhi::{
-    HostVulkanDevice, PresentFrame, VulkanAccess, VulkanGraphicsKernel, VulkanPresentTarget,
-    VulkanStage, VulkanTextureReadback,
+use streamlib_plugin_sdk::sdk::context::{
+    GpuContextFullAccess, GpuContextLimitedAccess, RuntimeContextFullAccess,
 };
-use streamlib::sdk::error::{Error, Result};
-use streamlib::sdk::rhi::{
+use streamlib_plugin_sdk::sdk::error::{Error, Result};
+use streamlib_plugin_sdk::sdk::rhi::{
     AttachmentFormats, ColorBlendState, ColorWriteMask, DepthStencilState, DrawCall,
     GraphicsBindingSpec, GraphicsDynamicState, GraphicsKernelDescriptor, GraphicsPipelineState,
     GraphicsPushConstants, GraphicsShaderStageFlags, GraphicsStage, MultisampleState,
-    PrimitiveTopology, RasterizationState, ScissorRect, Texture, TextureReadbackDescriptor,
-    TextureSourceLayout, VertexInputState, Viewport,
+    PresentTarget, PrimitiveTopology, RasterizationState, ScissorRect, Texture, TextureFormat,
+    TextureSourceLayout, VertexInputState, Viewport, VulkanAccess, VulkanGraphicsKernel,
+    VulkanLayout, VulkanStage,
 };
-use streamlib_consumer_rhi::VulkanLayout;
+use streamlib_plugin_abi::{ColorTraitsRepr, HdrStaticMetadataRepr, RawWindowHandleRepr};
 
+use raw_window_handle::{HasDisplayHandle, HasWindowHandle, RawDisplayHandle, RawWindowHandle};
 use winit::application::ApplicationHandler;
 use winit::dpi::PhysicalSize;
 use winit::event::WindowEvent;
@@ -41,13 +43,20 @@ pub struct LinuxWindowId(pub u64);
 
 static NEXT_WINDOW_ID: AtomicU64 = AtomicU64::new(1);
 
+/// Descriptor-set ring depth for the display blit kernel. The engine's
+/// `MAX_FRAMES_IN_FLIGHT` lives in engine-internal code an engine-free
+/// package cannot name, so the display owns its own frames-in-flight
+/// policy for the blit kernel's descriptor sets (matched to the
+/// swapchain's 2-deep present ring).
+const DISPLAY_BLIT_FRAMES_IN_FLIGHT: u32 = 2;
+
 /// Compiled-in display-blit SPIR-V (built by this package's `build.rs`).
 const DISPLAY_BLIT_VERT_SPV: &[u8] =
     include_bytes!(concat!(env!("OUT_DIR"), "/display_blit.vert.spv"));
 const DISPLAY_BLIT_FRAG_SPV: &[u8] =
     include_bytes!(concat!(env!("OUT_DIR"), "/display_blit.frag.spv"));
 
-#[streamlib::sdk::processor("Display")]
+#[streamlib_plugin_sdk::sdk::processor("Display")]
 pub struct LinuxDisplayProcessor {
     gpu_context: Option<GpuContextLimitedAccess>,
     window_id: LinuxWindowId,
@@ -61,7 +70,7 @@ pub struct LinuxDisplayProcessor {
     stop_called: Arc<AtomicBool>,
 }
 
-impl streamlib::sdk::processors::ManualProcessor for LinuxDisplayProcessor::Processor {
+impl streamlib_plugin_sdk::sdk::processors::ManualProcessor for LinuxDisplayProcessor::Processor {
     fn setup(&mut self, ctx: &RuntimeContextFullAccess<'_>) -> Result<()> {
         tracing::trace!("Display: setup() called");
         self.gpu_context = Some(ctx.gpu_limited_access().clone());
@@ -93,7 +102,7 @@ impl streamlib::sdk::processors::ManualProcessor for LinuxDisplayProcessor::Proc
         Ok(())
     }
 
-    fn start(&mut self, ctx: &RuntimeContextFullAccess<'_>) -> Result<()> {
+    fn start(&mut self, _ctx: &RuntimeContextFullAccess<'_>) -> Result<()> {
         tracing::trace!(
             "Display {}: start() called — spawning render thread",
             self.window_id.0
@@ -115,18 +124,14 @@ impl streamlib::sdk::processors::ManualProcessor for LinuxDisplayProcessor::Proc
             .clone()
             .unwrap_or(ScalingMode::Stretch);
 
+        // Engine-free GPU access: the render thread mints every GPU resource
+        // host-side via `gpu_context.escalate(|full| full.create_*(..))`, never
+        // a raw host device. `GpuContextLimitedAccess` is `Clone` + `Send`, so
+        // the clone rides into the render thread.
         let gpu_context = self
             .gpu_context
             .clone()
             .ok_or_else(|| Error::Configuration("GPU context not initialized".into()))?;
-
-        // Manual-mode start() takes FullAccess directly. For cdylib-
-        // resident processors the engine wraps lifecycle dispatch in
-        // `with_cdylib_scope` (#1072), so `ctx.gpu_full_access()`
-        // routes through the FullAccess vtable transparently — same
-        // contract as in-process. Calling `.escalate(...)` here would
-        // re-enter the gate held by the wrap and panic.
-        let vulkan_device = ctx.gpu_full_access().host_vulkan_device_arc()?;
 
         running.store(true, Ordering::Release);
 
@@ -198,13 +203,15 @@ impl streamlib::sdk::processors::ManualProcessor for LinuxDisplayProcessor::Proc
                             .ok()
                             .and_then(|s| s.parse::<u64>().ok())
                             .unwrap_or(30);
-                        // Test-only: when set, the render closure deliberately
+                        // Test-only: when set, the render body deliberately
                         // returns `Err` at the given frame counter, exercising
-                        // `VulkanPresentTarget`'s error-path drain (consumes the
-                        // `image_available_semaphore` so the next slot reuse
-                        // doesn't trip `VUID-vkQueueSubmit2-semaphore-03868`).
-                        // Run with `VK_LOADER_LAYERS_ENABLE=*validation*` to
-                        // surface any drain bug as a VUID at the next acquire.
+                        // the present target's error-path drain (the host
+                        // `end_frame` closes any dangling render pass and
+                        // consumes the `image_available_semaphore` so the next
+                        // slot reuse doesn't trip
+                        // `VUID-vkQueueSubmit2-semaphore-03868`). Run with
+                        // `VK_LOADER_LAYERS_ENABLE=*validation*` to surface any
+                        // drain bug as a VUID at the next acquire.
                         let inject_error_at_frame =
                             std::env::var("STREAMLIB_DISPLAY_INJECT_ERROR_AT_FRAME")
                                 .ok()
@@ -232,7 +239,6 @@ impl streamlib::sdk::processors::ManualProcessor for LinuxDisplayProcessor::Proc
 
                         let mut app = DisplayEventLoopHandler {
                             window: None,
-                            vulkan_device,
                             gpu_context,
                             inputs,
                             running,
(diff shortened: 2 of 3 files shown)


tato123:
## Verify verdict: PASS — with two documented review items (do NOT merge before the E2E, per owner directive)

**Headline soundness win (verify-confirmed):** the raw-`HostVulkanDevice` Arc transit is gone — every GPU resource is now minted host-side via `escalate(|full| full.create_present_target/create_graphics_kernel/create_texture_readback)`, eliminating the separately-built-`.slpkg` driver-corruption class (`docs/learnings/slpkg-raw-device-rhi-construction.md`). escalate placement + wire-reprs verified.

**should-fix 1 — dead-blit-shader exit criterion (resolved by documentation, not silent skip):** the grep required by the ticket's criterion 3 found the engine `runtime/streamlib-engine/src/vulkan/rhi/shaders/display_blit.{vert,frag}` are **production-dead but test-alive** — consumed by the engine's own graphics-kernel test fixtures (`vulkan_graphics_kernel.rs` `vert_spv()/frag_spv()`, ~L3155, inside `#[cfg(test)]`). So the engine copy is correctly **kept**, not deleted; criterion 3 resolves to 'not dead — engine test fixtures consume it.' A clean removal (migrating those fixtures to a synthetic SPIR-V) is a separate engine-test PR.

**should-fix 2 — camera→display E2E NOT YET RUN (the merge gate):** exit criterion 2 (separately-built `.slpkg` camera(vivid)→display, PNG check, headless-drain degradation, error-injection semaphore-drain) is unverified — it's the class of bug tier-2 hardware tests can't catch. **This PR is code-verified + engine-hardware-verified but not `.slpkg`-E2E-verified.** Standing up the autonomous `sg docker --gpus all` vivid→display E2E now; evidence attaches here before merge. Hold merge until then.

PR #1394, opened 2026-07-17
feat(pack): opt-in cargo source-mirror — precursor for the separate-build validation gate

tato123:
## What & why

The separate-build `.slpkg` validation gate (**Part of #1257**) must build a package the way a real downstream consumer would: resolving the engine crates **by bare version** (`streamlib = "0.6.0"`), **link-free** — no `streamlib link`, no `[patch.crates-io]`. A `streamlib link` build resolves the engine to the *same workspace source* via `[patch.crates-io]`, so its layout is matched to the host by construction and it **masks** the exact distribution hazards the gate exists to catch (raw-device double-free, ABI-skew refusal — see `docs/learnings/slpkg-raw-device-rhi-construction.md`). Nothing today resolves `streamlib = "0.6.0"` for a link-free build, so this adds the missing precursor.

**What:** `cargo xtask static-registry emit --cargo-mirror` additionally emits a cargo **source-replacement mirror** (`cargo vendor` of the full crates.io closure + the workspace engine/SDK crates injected byte-stable) plus a generated `[source.crates-io] replace-with` config. A package then resolves + compiles its whole closure **offline** from the tree. Opt-in (off by default) so a routine `.slpkg` emit doesn't pay for the ~690MB vendor.

**Not a return of the removed registry (#1322):** #1322 removed the SDK-distribution registry *emulation* (the never-deployed `registry.tatolab.com` sparse index + `registry use/serve` CLI). This is an emit-machine-**local**, gate-only affordance — a cargo directory source resolves by an absolute local path and can't be HTTP-served, so it's deliberately outside the relocatable tree. the ratified precedent for the `[source]`-replacement pattern is #1278; `crate_tarball.rs` returns verbatim as the #1250 byte-stability helper (structural `.crate` verification + vcs-info-strip normalization), not registry emulation.

**Alternatives rejected:** link/`[patch]` (layout-matched → masks the hazards; excluded by the gate's "no path/link overrides"); a second-checkout `[patch]` (still a path override, misses `.crate` packaging bugs); publishing to real crates.io (owner-gated release act, and the name is squatted — below).

## Validation (real output)

- Link-free offline resolve + **compile**: `streamlib = "0.6.0"` locks 521 packages; `cargo build --locked --offline` compiles the engine chain from the mirror.
- Byte-stable: injected `.crate` `package` checksum identical across independent re-emits; `--locked`-clean (canonical crates.io source id preserved).
- Non-vacuous negative control: `--offline` with no replacement fails (`location searched: crates.io index`).
- `tools/streamlib-pack/tests/cargo_mirror_offline_build.rs` (`--ignored`) encodes all of it; a new integration test covers the `--cargo-mirror` closure glue; the full gate battery is green (18/0).

## ⚠️ Squat finding (future decision — not forced here)

The `streamlib` crate name is **squatted on crates.io** (`streamlib = 0.6.4`, unrelated). Bare `^0.6.0` admits it, so an *online* resolve without the replacement would lock the squatter — but `streamlib-engine` / `-idents` / `-consumer-rhi` don't exist on crates.io, so an **offline** resolve fails hard and the replacement shadows the squatter (the test asserts `streamlib-engine` in the lock). The gate is offline-by-construction → no silent-wrong-crate risk. Whether to eventually **rename/reserve** the engine crate name is a strategic call tied to the future public-publish step — flagged, yours to make later.

## Scope

Precursor only — the gate ticket stays open. Its exit criteria (the one-command gate — emit→resolve→run→verdict — its use by a conversion, and the stale-ABI negative case) are untouched by the mirror alone. The release-completeness pin seam (`crates_missing_from_release`) is only half-reconnected (it keys on `registry=tatolab` pins that #1322 stripped) — noted as gate-work.

Part of #1257 · direction adjudicated + owner-greenlit (full-offline vendor).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5256,13 +5256,16 @@ name = "streamlib-pack"
 version = "0.6.0"
 dependencies = [
  "anyhow",
+ "flate2",
  "libc",
  "serde",
  "serde_json",
  "serde_yaml",
+ "sha2",
  "streamlib-cargo-build",
  "streamlib-idents",
  "streamlib-processor-schema",
+ "tar",
  "tempfile",
  "thiserror 2.0.18",
  "toml 0.8.23",
--- a/docs/architecture/package-development-model.md
+++ b/docs/architecture/package-development-model.md
@@ -272,8 +272,15 @@ workspace member ∧ linkable name (`streamlib*` / `vulkan-jpeg` /
 topological order. There is no "SDK-subset vs. all-libs" switch —
 `streamlib-plugin-sdk` / `vulkan-jpeg` are members by definition. It is the
 crate set `streamlib link` overrides (and the set a future SDK release would
-publish); the `.slpkg` emit no longer consults it (a `.slpkg` release
-publishes packages, not crates).
+publish). A routine `.slpkg` emit does not consult it (a `.slpkg` release
+publishes packages, not crates); a `--cargo-mirror` emit *does* — it records the
+closure in the release manifest's `crates` alongside the cargo source-replacement
+mirror it emits (see
+[`static-registry.md`](static-registry.md#the-opt-in-cargo-mirror)).
+
+> ~~The `.slpkg` emit no longer consults it.~~ — Superseded 2026-07-17 by the
+> separate-build validation gate's `--cargo-mirror` emit path, which consults
+> `compute_release_closure`. The claim held only for a routine (no-flag) emit.
 
 **Manifest-last atomicity.** A `ReleaseManifest`
 ([`sdk/streamlib-idents/src/release.rs`](../../sdk/streamlib-idents/src/release.rs))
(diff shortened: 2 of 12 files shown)


tato123:
Closing as premature (owner-directed, 2026-07-17). Publishing/consuming packages by version from a registry is a **future** task; today's model is dev-local consumption via `streamlib link` (a matched build), so the separate-build validation gate this mirror feeds (#1257) isn't a current need. This is part of re-scoping the milestone to **engine-free only** — the Gitea-era publishing assumption doesn't apply while consumption is link-based.

The cargo-mirror work (byte-stable `.crate` machinery, `[source.crates-io] replace-with` full-offline vendor, the crates.io squat finding) is preserved on branch `feat/engine-crate-mirror-1257` and in this PR's history for when publishing-to-cargo actually lands.

PR #1431, opened 2026-07-18
feat(plugin-sdk): dynamic Bag read/write over the msgpack wire

tato123:
Closes #1407

## Summary
- Adds the engine-free `Bag` type in `sdk/streamlib-plugin-sdk`: an eagerly-decoded msgpack named-map value tree with tolerant typed `get`/`get_opt`/`get_bin`, `set`/`set_bin`, and a `bag!` literal macro. Backed by `rmpv` so `bin` stays a distinct wire type from `array`.
- Adds a named error taxonomy in `streamlib-error`: `BagKeyMissing`, `BagTypeMismatch`, `BagDecodeFailed`, `BagEncodeFailed`, `BagPayloadOverflow` — no `()` errors, no panics.
- Adds `read_bag(port)` / `write_bag(port, &bag, ts)` wrapping the existing `read_raw`/`write_raw` (both raw methods stay public). `write_bag` enforces the design §11 phase-1 interim per-channel payload budget (`BAG_MAX_PAYLOAD_BYTES = 65536`, mirroring the host's `MAX_PAYLOAD_SIZE`) and surfaces overflow as `Error::BagPayloadOverflow`, never a panic/SIGSEGV. This budget is explicitly interim — #1421 (hint-primed dynamic slot allocation) removes it.
- No wire-format, vtable, or `#[repr(C)]` change — every payload was already a self-describing msgpack named map; this just makes it a first-class SDK type instead of raw bytes.
- Commits a canonical value-class-complete msgpack fixture (Rust is the source of truth, `sdk/streamlib-plugin-sdk/tests/fixtures/bag_conformance.msgpack`) and reads it identically from Rust `Bag`, Python `dict`, and Deno `object`, asserting every value class plus tolerant missing/wrong-typed/unexpected-field handling per runtime.
- Adds the eager-Bag-decode vs codegen-deserialize criterion microbench (`benches/bag_decode.rs`) per the hot-path rule.

## Test evidence
Run in-worktree on this branch (`4c6cf65c`):
- `cargo test -p streamlib-plugin-sdk -p streamlib-error --lib` → 92 passed, 0 failed, 1 ignored
- `cargo test -p streamlib-plugin-sdk --lib bag` → 8/8 `bag::tests::*` passed
- `cargo test -p streamlib-plugin-sdk --test bag_conformance` → 3 passed (`canonical_bag_is_byte_stable_against_fixture`, `tolerant_reads_of_missing_and_unexpected_fields`, `fixture_decodes_to_canonical_values`); `regenerate_fixture` intentionally `ignore`d (explicit-run-only fixture writer)
- `deno test --allow-read bag_conformance_test.ts` (sdk/streamlib-deno) → 3/3 passed
- `python3 -m pytest python/streamlib/tests/test_bag_conformance.py -v` (sdk/streamlib-python) → 3/3 passed

Rust ↔ Python ↔ Deno all decode the identical committed fixture to the same canonical values and agree on tolerant handling of missing/unexpected fields — cross-language wire disagreement would fail these tests.

These survived an independent verify pass; none blocked the PASS verdict, but flagging for owner visibility before merge:

1. **should-fix** — `sdk/streamlib-plugin-sdk/src/iceoryx2.rs:246`: Ticket §11 landmine coverage requires a schema-free bag that exceeds the fixed-slot budget to surface as the named `Error::BagPayloadOverflow` ("never a panic/SIGSEGV"). `write_bag` implements the guard (`data.len() > BAG_MAX_PAYLOAD_BYTES → BagPayloadOverflow`), and the guard is correct by inspection (constant 65536 == host `MAX_PAYLOAD_SIZE`; `>` correctly admits an exactly-full 65536-byte payload), but `grep -rn 'BagPayloadOverflow|write_bag|BAG_MAX_PAYLOAD'` over the whole worktree shows zero tests exercise the overflow branch. `write_bag` can't be unit-tested as-is because `OutputWriter` is only constructible from a live ABI vtable (`from_raw_parts`). Suggested: extract a pure `check_bag_payload_within_budget(port, encoded_len) -> Result<()>` helper that `write_bag` calls, and add a unit test asserting a >65536-byte encoding returns `Error::BagPayloadOverflow` and an ==65536 one does not. At minimum, note the missing gate test as a known review item (this item does that).

2. **should-fix** — `sdk/streamlib-plugin-sdk/src/bag.rs:112`: the `rmpv::ext::from_value(...).map_err(|e| Error::BagTypeMismatch { ... })` block is byte-for-byte identical at two sites — `get()` (L112-116) and `get_opt()` (L127-133) — real duplication; any change to how a type-mismatch is decoded/reported must be made in both places in lockstep. Suggested: extract a private `fn decode_value<T: DeserializeOwned>(&self, key: &str, value: &Value) -> Result<T>` that owns the `from_value` + `BagTypeMismatch` logic; `get` and `get_opt` both collapse to one line built on it.

3. **should-fix** — `sdk/streamlib-plugin-sdk/src/bag.rs:93`: `to_msgpack()` builds an intermediate `Value::Map` by deep-cloning every value (`v.clone()`) and allocating a new key `String` per entry (L91-95), only to hand the temporary map to `write_value` and immediately drop it — a full-tree clone of the whole bag on every encode, and `to_msgpack()` is on the documented write hot path (`write_bag`), which this PR ships a bench for. Suggested: stream directly with no clone — `rmp::encode::write_map_len` + per-entry `rmp::encode::write_str` + `rmpv::encode::write_value(&mut out, v)` (already takes `&Value`, so values serialize by reference). Pulls the sibling `rmp` crate as a direct dep (already transitive via `rmp-serde`/`rmpv`) — a fair tradeoff for a hot-path encoder, or accept the clone as a stated cost for tiny bags.

4. **info** — `sdk/streamlib-plugin-sdk/src/iceoryx2.rs:52`: `BAG_MAX_PAYLOAD_BYTES = 65536` hardcodes the host's `streamlib_ipc_types::MAX_PAYLOAD_SIZE` across the ABI boundary with no compile-time link, so the two silently drift if the host slot size ever changes. Acknowledged/documented as interim (#1421); the engine-free SDK genuinely can't dep the host IPC crate. No action needed now — ensure #1421 removes the constant rather than leaving the un-linked twin permanently.

5. **info** — `sdk/streamlib-plugin-sdk/src/iceoryx2.rs:238`: `read_bag`/`write_bag` (the shipped IPC wrapper methods) have no in-process test — they require a live host/iceoryx2 vtable the sandbox can't observe (exit 144). Cross-language wire correctness is covered fixture-based, and the wrappers are thin (`to_msgpack` + budget + `write_raw`; `read_raw` + `from_msgpack`). Defer the round-trip-over-real-IPC check to `/verify-live` and treat as a rig-gated follow-up.

6. **info** — `sdk/streamlib-plugin-sdk/src/bag.rs:244`: the exported `bag!` macro expands to `Result::expect(...)` — panics at the construction site on a non-serializable literal. Deliberately modeled on `serde_json::json!`, documented as such, and only fires on a compile-site literal the author controls — defensible design, flagged only so it's a conscious choice (json!-parity is the right call; the panic path is literal-construction-only, not a runtime data path).

7. **low** — `sdk/streamlib-plugin-sdk/src/bag.rs:112`: `get()`/`get_opt()` `value.clone()` clones the stored `Value` subtree before `from_value` consumes it. Forced by `rmpv`'s owned-`Value` `from_value` API (no borrow-deserialize exists) — inherent to the eager-decode design, not a defect. No action required; if read-side allocation ever shows up hot, a borrowing accessor over `&Value` for scalar types could bypass the clone, but that's out of scope here.

## Scope
No wire-format changes, no new `#[repr(C)]` types/vtable slots, no config-as-bag, no channel naming — matches the ticket's stated out-of-scope list.


tato123:
Post-verify fix round applied (4 should-fix findings from the change-verifier + rust-craftsmanship lens):
- **bag.rs DRY** — extracted `decode_value<T>` so the `from_value` + `BagTypeMismatch` construction lives once (was duplicated verbatim in `get()`/`get_opt()`).
- **bag.rs hot-path** — `to_msgpack()` now streams the map zero-clone (`write_map_len` + per-entry `write_str`/`write_value` by reference) instead of deep-cloning the whole bag on every encode — the write path this PR benchmarks.
- **iceoryx2.rs** — extracted a pure `check_bag_payload_within_budget` helper and added tests for the §11 `BagPayloadOverflow` gate: `payload_at_budget_boundary_is_accepted` (==65536 ok) and `payload_over_budget_is_named_overflow` (>65536 → named error, never a panic).
- Dropped the unused `serde_bytes` dev-dep.

Full local gate battery green: **32 passed, 0 failed** — plugin-sdk lib (94 passed), bag_conformance (Rust/Python/Deno all decode the same Rust-authored fixture), bench compiles, clippy 0 errors, all xtask lints + license + boundary checks. Verify verdict was PASS; this round only polishes.

Ready for review + merge. This is issue A (#1407), the root of the M34 chain — the rest (#1408 → …) unblocks once it merges.

Issue #34, opened 2025-10-28
[Future] Command & Control (C2) coordinator and multi-unit operations


tato123:
## Overview
Design and implement Command & Control system for coordinating multiple power armor units in the field. C2 acts as strategic coordinator that assigns high-level goals while agents autonomously handle tactics.

**Status:** Planning phase - requires more design work and depends on A2A (#33)

## Vision
**"Radio in from C2, agents inside suits handle execution"**

```
C2 Coordinator: "Unit 1, clear building. Unit 2, provide overwatch."
  ↓
Agents receive goals, plan autonomously
  ↓
Agents coordinate via A2A
  ↓
Agents execute, report back
  ↓
C2 observes, provides strategic guidance
```

**Like a video game:** Commander gives orders, AI teammates execute intelligently.

### Mode 1: Observer (Passive)
**C2 watches but doesn't command**
- Listens to all A2A traffic
- Records mission data (feeds, telemetry, logs)
- Visualizes unit positions and status
- Human operator monitors
- Can intervene in emergencies only

**Use case:** Training, testing, observation

### Mode 2: Peer Agent (Collaborative)
**C2 is just another agent in the hive**
- Has different tools (map, intel database, communications)
- Suggests strategies: "Building has 3 exits, recommend flanking"
- Agents can accept or ignore suggestions
- No special authority, democratic coordination

**Use case:** Collaborative operations, equal partnership

### Mode 3: Coordinator (Default - Recommended)
**C2 assigns high-level goals, agents handle tactics**
- Radio voice: "Alpha team, clear building A"
- Agents hear goal, plan autonomously
- Agents coordinate via A2A to execute
- Agents report: "Building cleared" or "Need assistance"
- C2 provides strategic updates: "Intel reports hostiles in basement"

**Use case:** Mission operations, matches military doctrine

### 1. C2 Dashboard (Web UI)
**Views:**
- **Grid view**: Live camera feeds from all units (4x4 grid)
- **Map view**: GPS positions, waypoints, objectives
- **Telemetry**: Battery, temperature, bandwidth, status
- **Timeline**: Mission events, A2A messages, key moments
- **Agent status**: What each agent is currently doing

**Controls:**
- Voice input: Speak goals to units
- Manual override: Direct control if agent fails
- Mission planning: Set waypoints, objectives
- Recording: Start/stop mission recording

### 2. C2 Agent Architecture
**Option A: Embedded in C2 Runtime**
```rust
C2Runtime::new()
[...]

PR #1433, opened 2026-07-19
feat(plugin-sdk): config-as-bag

tato123:
## Closes #1408 — config-as-bag

Design: #1345 §11. Processor config today requires a mandatory config schema, and an empty schema degrades the typed config to `Option<Value>` — a footgun. With `Bag` landed, config is just a bag: dynamic by default, typed config is tolerant-deserialize sugar over it.

### What changed
- `Bag` becomes a first-class serde `Config`: no-config processors receive it directly instead of a schema-driven placeholder.
- Deleted the empty-schema → `Option<Value>` footgun in the JTD codegen (`sdk/streamlib-jtd-codegen/src/lib.rs`); an empty `config:` schema is now a hard `anyhow::bail!` that tells the author to drop the block entirely and fall through to `EmptyConfig`.
- `deny_unknown_fields` is banned on generated `*Config` types (config schemas stay tolerant-deserialize); escalate request/response schemas are untouched and keep strict deserialize, per the Design §11 "escalate stays typed" landmine.
- `EmptyConfig` gets hand-rolled `Serialize`/`Deserialize` impls (in both the engine core and the plugin SDK) so it tolerates an incoming `{}` map on the wire instead of rejecting it — a real correctness fix: the previous `#[derive(Deserialize)]` unit struct rejected a msgpack/JSON map shape, and under config-as-bag a no-config processor can receive one.
- New ABI tests: no-config, dynamic-bag config, and typed-view config all construct; extra unknown fields still deserialize; a malformed required field is a named error.
- Re-pinned `twin_drift_guard.rs`'s `processor_vtable` trip-wire after the config-delivery test additions.

### Test evidence
- `cargo test -p streamlib-plugin-sdk -p streamlib-jtd-codegen --lib` — 58 passed (jtd-codegen), 102 passed / 1 ignored (plugin-sdk), 0 failed.
- `cargo test -p streamlib-engine --lib -- core::plugin::twin_drift_guard` — 3 passed, 0 failed.
- `cargo test -p streamlib-engine --test attribute_macro_test` — 8 passed, 0 failed, including `empty_config_is_a_tolerant_bag`.

### Verification
Independent verify-change (change-verifier + path-routed domain lenses) ran clean on this branch; all lenses cleared.

### Review items (non-blocking)
These surfaced during verification and don't block merge, but are worth the owner's eyes:

1. **should-fix** — `packages/h264/schemas/h264_decoder_config.yaml:9`: Ticket AC says "Canonical-pattern sweep: every in-repo processor on the old config path (packages/, examples/, Python, Deno) migrates in the same PR." The diff touches zero package/example/Python/Deno files. Four in-repo processors still carry the old empty-schema workaround (`optionalProperties: {}` / a placeholder-only body) specifically to dodge the `Option<Value>` footgun: `packages/h264/schemas/h264_decoder_config.yaml`, `packages/h265/schemas/h265_decoder_config.yaml`, `packages/jpeg/schemas/jpeg_decoder_config.yaml`, `packages/debug-utilities/schemas/video_frame_counter_config.yaml`. Their explanatory comments (e.g. "an empty schema becomes `pub type X = Option<Value>`") now describe behavior this PR turned into a hard `anyhow::bail!` error; the new bail message itself tells authors to drop the `config:` block and fall to `EmptyConfig` — the exact migration these files did not receive. They still compile (empty struct, no build break) and are now tolerant, but they are living contradictions of the new canonical pattern. Milestone #34's body explicitly defers package modernization ("existing packages modernize as part of their Engine-Free Package Migration conversion — convert once, modernize once"), which sanctions the deferral over the ticket's "same PR" wording. Ship it, but state this deferral (and the now-stale schema comments) explicitly here so the sweep is tracked, not silently dropped.

2. **should-fix** — `sdk/streamlib-plugin-sdk/src/processors.rs:89`: `EmptyConfig`'s hand-rolled `Serialize`/`Deserialize` impls are byte-identical twins in the SDK (`sdk/streamlib-plugin-sdk/src/processors.rs:89-105`) and the engine (`runtime/streamlib-engine/src/core/processors/mod.rs:43-59`), and they are wire-load-bearing — config crosses the plugin ABI, so both sides must serialize to the same empty-named-map shape — yet no drift guard covers them. `twin_drift_guard.rs` guards only the `plugin/` marshalling files (`forwarding_subscriber`, `layout_skew_diagnostic`, `processor_vtable`), NOT the `processors` modules. This diff converts `EmptyConfig` from a trivially-self-evident `derive` into 17 lines of hand-written serde that can now silently diverge: change the SDK side's shape (or the engine's) without the other and config decode breaks across the ABI with no failing test. Both files carry the identical block: `serde::ser::SerializeMap::end(serializer.serialize_map(Some(0))?)` for `Serialize` and `deserializer.deserialize_ignored_any(serde::de::IgnoredAny)?; Ok(EmptyConfig)` for `Deserialize`. Suggested next step: add a logic-identical drift-guard entry for the `EmptyConfig` serde twins (or host `EmptyConfig` in a shared engine-free crate both import), so a one-sided edit to the wire shape fails `cargo test --lib` rather than shipping in a `.slpkg`.

3. **low** — `sdk/streamlib-jtd-codegen/src/lib.rs:823`: The empty-schema guard re-allocates an invariant `String` on every line of the codegen output: `format!("pub type {ROOT_NAME_SENTINEL}")` sits inside the `.any(|line| ...)` closure, so it is rebuilt per line even though `ROOT_NAME_SENTINEL` is a const. Suggested next step: bind `let type_alias_prefix = format!("pub type {ROOT_NAME_SENTINEL}");` once above the `.any` and borrow it in the closure. Low because this is a build-time codegen tool over small inputs, not a hot path.

4. **low** — `runtime/streamlib-engine/src/core/processors/mod.rs:48`: `EmptyConfig`'s `Serialize` uses the UFCS form `serde::ser::SerializeMap::end(serializer.serialize_map(Some(0))?)`, which reads less clearly than the method-chain form `serializer.serialize_map(Some(0))?.end()` used for `Bag` in `bag.rs`. In the engine copy `SerializeMap` is not imported so UFCS is required, but the SDK copy (`processors.rs:94`) mirrors the same UFCS despite the SDK crate otherwise preferring the method form — a small cross-file readability inconsistency. Optional: in the SDK copy bring `SerializeMap` into scope and use `serializer.serialize_map(Some(0))?.end()` to match the `Bag` impl's style; leave the engine copy as-is if the trait isn't imported there.

5. **info** — `sdk/streamlib-jtd-codegen/src/lib.rs:837`: `is_config_schema = expected_struct_name.ends_with("Config")` decides which generated types lose `deny_unknown_fields`. The gate is a name-suffix heuristic. It is correct for the current schema set — escalate types are `EscalateRequest`/`EscalateResponse` (verified they keep strictness), and every `*Config` is a processor config — so the Design §11 "escalate stays typed" landmine is respected today. But any future non-config wire type named `*Config` would silently drop strict deserialize with no test catching it. No action required; if the config/non-config distinction ever needs to be firmer, thread an explicit schema-kind flag from the caller instead of the name suffix.

6. **info** — `runtime/streamlib-engine/tests/attribute_macro_test.rs:84`: The tolerant `EmptyConfig` serde impls are a real correctness fix, not just cosmetic. On main, `#[derive(Deserialize)] struct EmptyConfig;` (unit struct) rejects a msgpack/JSON map wire shape ("invalid type: map, expected unit struct"). Under config-as-bag a no-config processor can receive `{}` on the wire. The custom impls + `empty_config_is_a_tolerant_bag` and the `no_config_processor_constructs_and_ignores_unexpected_fields` construct-slot test are non-vacuous: reverting the impls (unit-struct derive) fails both. Twin `EmptyConfig` copies (engine `core/processors/mod.rs` and SDK `processors.rs`) were updated identically.

Closes #1408.


tato123:
**Scope note — canonical-pattern sweep, sanctioned deferral (tracked, not dropped):**

Four in-repo processors still carry the old empty-schema workaround (`optionalProperties: {}` placeholder bodies that existed only to dodge the `Option<Value>` footgun this PR deletes):
- `packages/jpeg/schemas/jpeg_decoder_config.yaml` (note: jpeg's decoder genuinely consumes `max_width`/`max_height`, so its config is real — only the empty-schema comment is stale)
- `packages/h264/schemas/h264_decoder_config.yaml`
- `packages/h265/schemas/h265_decoder_config.yaml`
- `packages/debug-utilities/schemas/video_frame_counter_config.yaml`

They still compile and are now tolerant (empty struct, no `Option<Value>`), so nothing is broken — but their explanatory comments now describe behavior that became a hard `bail!` error, and h264/h265/debug-utilities are standalone (non-workspace) packages that resolve deps via `streamlib link` and cannot build/verify in-sandbox. Converting them here would be an unverifiable commit.

Per the M34 milestone body — "existing packages modernize as part of their Engine-Free Package Migration conversion (convert once, modernize once)" — each package drops its stale `config:` block + comment during its own migration, not in this SDK-surface PR. Flagging explicitly so the sweep is tracked to that sequencing rather than assumed complete.

tato123:
Post-verify fix round applied (4 findings from the change-verifier + rust-craftsmanship lens), gate battery green:
- **EmptyConfig serde-twin drift guard** (should-fix) — the hand-rolled `Serialize`/`Deserialize` for `EmptyConfig` are byte-identical, wire-load-bearing twins in the SDK (`processors.rs`) and engine (`core/processors/mod.rs`); config crosses the plugin ABI, so a one-sided edit would break config decode with no failing test. Added a `twin_drift_guard` entry — `divergent_empty_config_serde_twin_is_tripwired` now trips on a one-sided change.
- **Escalate-strictness trip-wire** (§11 #7) — `post_process_rust_escalate_schemas_stay_strict_config_suffix_drops_strict` locks the contract that escalate request/response schemas keep `deny_unknown_fields` while `*Config` schemas drop it, so the name-suffix heuristic can't silently regress.
- Hoisted a per-line `format!` in the codegen empty-schema guard; switched the SDK `EmptyConfig::serialize` to the method-chain form to match `Bag`.

Local gate battery: **28 passed, 0 failed** — plugin-sdk lib (102), engine lib (1628), all 4 `twin_drift_guard` tests, jtd-codegen (59), attribute_macro integration, clippy 0 errors, full xtask lint suite + license + boundaries. Verify verdict was PASS; this round only hardens.

Ready for review + merge. Canonical-sweep deferral tracked in the scope note above. This is issue B (#1408); #1409 (ports-in-code) unblocks on merge.

PR #1441, opened 2026-07-19
feat(engine): Runner::add_local — register a #[processor] host type live

tato123:
Closes #1410

## Summary

Adds `Runner::add_local::<P>(config: serde_json::Value) -> AddedModule`, a live-registration op that registers a `#[processor]`-annotated host type directly onto the existing `PROCESSOR_REGISTRY` — no package on disk, no dylib. Packaging stays for *sharing* a processor; it is never a prerequisite to *having* one.

- Reuses the module_loader's transactional staging/ledger/rollback seam via a new session module-source variant (`session.rs`, `staging.rs`) — no parallel local registry.
- Mints runtime identities under a reserved `@session/<kebab-name>@0.0.N` org (`streamlib-idents`), with a monotonically increasing version per re-registration.
- `@session` is refused at `pkg build` / registry publish (`tools/streamlib-pack`) — session names can never masquerade as an installable package.
- Local-vs-installed same-name collision **warns** and keeps both addressable (never silently overwrites).
- Named error variants for duplicate `@session` name and malformed name (`module_loader/errors.rs`).
- Config is validated against `P::Config` (JSON round-trip, same currency as `ProcessorSpec.config`) before registration.

Landed in 3 commits:
- `bbeab4ac` feat(idents): reserve `@session` org — non-publishable, refused at pkg build
- `d4d29480` feat(idents): mint `@session/<name>@0.0.N` runtime identities
- `f51d795e` feat(engine): `Runner::add_local` — register a `#[processor]` host type live

## Test evidence

All commands run in-worktree against `f51d795e` on this branch:

```
cargo test -p streamlib-engine --lib module_loader::session
  test core::runtime::module_loader::session::kebab_tests::kebabs_pascal_case_type_names ... ok
  test core::runtime::module_loader::session_tests::add_local_accepts_a_well_formed_config ... ok
  test core::runtime::module_loader::session_tests::add_local_refuses_a_duplicate_live_session_name ... ok
  test core::runtime::module_loader::session_tests::add_local_re_registers_after_removal_at_a_fresh_version ... ok
  test core::runtime::module_loader::session_tests::add_local_rejects_a_config_that_does_not_match_the_type ... ok
  test core::runtime::module_loader::session_tests::add_local_registers_a_session_processor_with_no_package_on_disk ... ok
  test core::runtime::module_loader::session_tests::add_local_shadow_keeps_both_installed_and_session_addressable ... ok
  test result: ok. 7 passed; 0 failed

cargo test -p streamlib-idents session
  test ident::tests::session_org_stays_constructible_via_grammar ... ok
  test ident::tests::is_reserved_for_session_flags_only_session_org ... ok
  test session::tests::mint_rejects_a_malformed_name ... ok
  test session::tests::mint_stamps_a_monotonically_increasing_version ... ok
  test session::tests::minted_module_ident_renders_under_the_session_org ... ok
  test result: ok. 5 passed; 0 failed

cargo test -p streamlib-pack --lib slpkg_assembly_rejects_reserved_session_org
  test tests::slpkg_assembly_rejects_reserved_session_org_but_staged_dir_is_exempt ... ok
  test result: ok. 1 passed; 0 failed

cargo test -p streamlib-pack   (full crate, sanity)
  3 + 1 + 4 unit/integration tests across catalog/registry/slpkg suites ... all ok
```

No GPU/IPC runtime surface touched (registry + registration bookkeeping only) — no live-rig E2E needed for this change.

## Acceptance criteria (from #1410)

- [x] `add_local::<T>()` registers on the existing `PROCESSOR_REGISTRY` and returns `AddedModule` (same shape `add` returns)
- [x] Runtime-dynamic op reusing module_loader's staging/ledger/rollback seam (session module-source variant) — verified against a constructed runtime; see review item below on the running-runtime leg
- [x] `@session` org reserved in `streamlib-idents` + refused at `pkg build`
- [x] Local-vs-installed same-name collision warns, both addressable
- [x] Named error variants (duplicate `@session` name, malformed name)
- [x] Tests: register+instantiate with no package on disk; collision warns + both resolvable; `@session/` rejected at `pkg build`

### Signature deviation from the 2026-07-18 acceptance-criteria correction

The correction says `add_local::<T>(bag)` taking a config `Bag` (default `Bag::empty()`). The implementation is `add_local::<P>(config: serde_json::Value)`. `Bag` lives in the engine-free `sdk/streamlib-plugin-sdk`, which the engine cannot depend on without inverting layering, and the `Config` trait mandates a JSON round-trip — `serde_json::Value` is the same currency `ProcessorSpec.config` already uses. The Bag ergonomic surface is scoped to #1412's App sugar (`sdk/streamlib-sdk`), which converts Bag→json before calling this primitive. Flagged as a review item below for the owner to confirm the layering call.

These surfaced during verification and don't block merge; listed here so they're visible in-context for the owner.

1. **should-fix** — `runtime/streamlib-engine/src/core/runtime/module_loader/session.rs:70`: The 2026-07-18 acceptance-criteria correction says the signature is `add_local::<T>(bag)` taking a config `Bag` (default `Bag::empty()`); the implementation is `add_local<P>(config: serde_json::Value)`.
   - Evidence: session.rs:70 `pub fn add_local<P>(&self, config: serde_json::Value) -> AddedModule`. Bag lives in the engine-free `sdk/streamlib-plugin-sdk` (bag.rs:44), which the engine cannot depend on without inverting layering; and the `Config` trait (core/processors/traits/config.rs) mandates JSON round-trip, so `serde_json::Value` — the same currency as `ProcessorSpec.config` — is the consistent engine-layer choice. The Bag ergonomic surface belongs to #1412's App sugar (`sdk/streamlib-sdk`), which can convert Bag→json.
   - Suggested next step: Document the deviation on the PR body (engine primitive takes serde_json::Value; Bag surface deferred to #1412 App sugar which bridges Bag→json). No code change needed if the owner accepts the layering rationale.

2. **should-fix** — `runtime/streamlib-engine/src/core/runtime/module_loader/session_tests.rs:170`: Acceptance criterion 2 requires the op be valid against a constructed AND a running runtime; tests only exercise a freshly-constructed `Runner::new()` (never started).
   - Evidence: Every test builds `Runner::new().unwrap()` and calls `add_local_blocking` without starting the runtime. No test drives `add_local` (async) or exercises registration after the runtime is running. The mechanism (spawn_blocking + PROCESSOR_REGISTRY/ledger under MODULE_REGISTRY_COMMIT_LOCK) is runtime-state-independent, so this is a coverage gap, not a demonstrated defect.
   - Suggested next step: Add a test that registers via add_local after the runtime is started, or file a rig-gated follow-up if starting a full runtime requires GPU/IPC unavailable in the sandbox (exit 144). Note the gap on the PR.

3. **info** — `runtime/streamlib-engine/src/core/runtime/module_loader/session.rs:175`: `config` is validated against `P::Config` then discarded — the registration stores only the descriptor + vtable, never the config.
   - Evidence: prepare_session_registration does `serde_json::from_value::<P::Config>(config)` purely for validation; the config value is dropped and the registered `ProcessorDescriptor` carries no config. A processor instance's real config is supplied later at add_processor time. Early validation can drift from the config actually used to instantiate.
   - Suggested next step: Confirm with #1412 App-sugar design that the add_local config param is intentionally validate-only (not instance-binding); the returned AddedModule is a module handle, not an instance handle.

4. **low** — `runtime/streamlib-engine/src/core/runtime/module_loader/staging.rs:540`: The `Err(e)` arm of `register_via_vtable` in commit_session_processor_registration logs a bug-grade error but then still inserts a ledger record and returns `Ok(processor_ident)`, reporting false success.
   - Evidence: register_via_vtable (processor_instance_factory.rs:882) currently returns `Ok(())` on every path (dedups duplicates by returning Ok), so the Err arm is effectively dead defensive code. If register_via_vtable ever gains a real Err path, the current shape would swallow it and return a spurious success.
   - Suggested next step: If register_via_vtable stays infallible, this is harmless; otherwise propagate the error instead of logging-and-inserting an empty ledger record.

5. **low** — `runtime/streamlib-engine/src/core/runtime/module_loader/session.rs:257`: `add_local_blocking`'s `unreachable!()` for the guarded `ExternalTokioHandle` arm omits the message string its sibling uses, a small consistency gap.
   - Evidence: Line 257 is `TokioRuntimeVariant::ExternalTokioHandle(_) => unreachable!(),` with a `// Guarded above` comment. The canonical siblings in the same module write `unreachable!("guarded above")` (mod.rs:435) — the guard-then-rematch-with-unreachable shape is the established convention here, so the code is structurally correct; only the panic message is missing.
   - Suggested next step: Match the sibling: `unreachable!("guarded above")` so a would-be panic (a real bug) names why it was thought impossible. Do NOT restructure into a single match — the two-step guard-then-match is the file's canonical pattern (mod.rs:422/435, 463/469) and consistency wins.

6. **low** — `runtime/streamlib-engine/src/core/runtime/module_loader/session.rs:404`: `kebab_case` reimplements PascalCase word-boundary splitting that already exists as `to_snake_case` and `pascal_to_snake`.
   - Evidence: `to_snake_case` (sdk/streamlib-processor-schema/src/processor_schema.rs:689, pub) and `pascal_to_snake` (sdk/streamlib-jtd-codegen/src/lib.rs:1687) already walk the same upper/lower/digit boundaries; for every case in `kebab_tests` the output equals `to_snake_case(name).replace('_', "-")`. This is a 4th hand-rolled case-splitter. It stays a *low* because it has a single call site (below engine-doctrine's ">1 call site" bar for mandatory extraction) and no *kebab* converter exists today — the extra dedup/trim logic is genuinely new.
   - Suggested next step: Optional: derive from the canonical splitter (`to_snake_case(type_name).replace('_', "-")` plus the existing trim/dedup), or leave as-is and note in the PR body that a shared case-conversion util is the eventual home. No change required to merge.

7. **info** — `runtime/streamlib-engine/src/core/runtime/module_loader/session.rs:396`: `expect()` appears in library code in `fallback_session_module` and (in the idents crate) `mint_session_module_ident`/`Org::new(SESSION_ORG)`.
   - Evidence: session.rs:396-397 `mint_session_module_ident("local").expect("`local` is a valid session package name")` and sdk/streamlib-idents/src/session.rs:699 `Org::new(SESSION_ORG).expect("SESSION_ORG passes the org grammar by construction")`. Both are infallible-by-construction (a hardcoded grammar-valid constant) with descriptive messages, which is the accepted idiom for a true invariant rather than a `?`-able error — not a defect, noted only because the repo mandates `?` over unwrap/expect broadly.
   - Suggested next step: No action. These are correct uses of `expect` on compile-constant invariants; leaving them is preferable to inventing an impossible error variant.

8. **info** — `runtime/streamlib-engine/src/core/runtime/module_loader/staging.rs:499`: The staged-processor drain/register loop in `commit_session_processor_registration` parallels the existing loop in `commit_module_load_registrations`.
   - Evidence: staging.rs:499-529 drains `staging.processors` and dispatches on `StagedProcessorRegistrationKind` exactly like staging.rs:418-452. The two intentionally diverge (session collects a flat `Vec<SchemaIdent>` and treats the `Dynamic` arm as bug-grade; the module-load path groups by owner and registers `Dynamic` normally), so this is incidental similarity of two loops that will evolve independently, not extract-worthy duplication.
   - Suggested next step: Leave as-is; the semantic divergence (Dynamic-is-a-bug vs Dynamic-is-valid, flat vs by-owner) makes a shared helper a net loss. Nothing to change.

9. **info** — `runtime/streamlib-engine/src/core/processors/processor_instance_factory.rs:16`: `warn_on_short_name_shadow` returns `Vec<SchemaIdent>` whose only consumer is tests — the sole production caller ignores it.
   - Evidence: The prod call at staging.rs:488 is `PROCESSOR_REGISTRY.warn_on_short_name_shadow(&processor_ident);` with the return value discarded; the doc (line 15) states the Vec is "for tests / diagnostics." Returning data purely to give tests a hook is a mild smell, but it doubles as a legitimate diagnostic surface and costs one allocation only on the (rare) collision path.
   - Suggested next step: Acceptable as diagnostics. If tightening later, the collision set could be asserted via a `tracing`-capture test instead, but no change is needed to merge.

### DIFF
--- a/runtime/streamlib-engine/src/core/processors/processor_instance_factory.rs
+++ b/runtime/streamlib-engine/src/core/processors/processor_instance_factory.rs
@@ -1276,6 +1276,38 @@ impl ProcessorInstanceFactory {
         self.descriptors.read().get(processor_type).cloned()
     }
 
+    /// Warn when `ident`'s short type name collides with an already-registered
+    /// processor under a different `(org, package)` — the session-local vs
+    /// installed name-shadow case. Both stay addressable by full ident (the
+    /// registry keys on the structured `SchemaIdent`, never the short name), so
+    /// this only surfaces the ambiguity a bare short-name reference would hit;
+    /// it never removes or overwrites either registration. Returns the shadowed
+    /// idents (for tests / diagnostics).
+    pub fn warn_on_short_name_shadow(&self, ident: &SchemaIdent) -> Vec<SchemaIdent> {
+        let shadowed: Vec<SchemaIdent> = self
+            .descriptors
+            .read()
+            .keys()
+            .filter(|other| {
+                other.r#type == ident.r#type
+                    && (other.org != ident.org || other.package != ident.package)
+            })
+            .cloned()
+            .collect();
+        for other in &shadowed {
+            tracing::warn!(
+                registering = %ident,
+                shadows = %other,
+                "processor short type name '{}' is now registered under two \
+                 distinct packages; both remain addressable by their full \
+                 `@org/package/Type@version` ident, but a bare short-name \
+                 reference is ambiguous",
+                ident.r#type,
+            );
+        }
+        shadowed
+    }
+
     /// List all registered processor types with their full descriptors.
     pub fn list_registered(&self) -> Vec<ProcessorDescriptor> {
         self.descriptors.read().values().cloned().collect()
--- a/runtime/streamlib-engine/src/core/runtime/module_loader/errors.rs
+++ b/runtime/streamlib-engine/src/core/runtime/module_loader/errors.rs
@@ -438,6 +438,59 @@ pub enum AddModuleError {
         package: streamlib_idents::PackageRef,
         processor_type: crate::core::descriptors::SchemaIdent,
     },
+
+    /// [`Runner::add_local`] was handed a type whose
+    /// [`GeneratedProcessor::descriptor`] returned `None` — it carries no
+    /// registerable descriptor, so there is nothing to register under
+    /// `@session/…`. Annotate the type with `#[processor(...)]`.
+    ///
+    /// [`Runner::add_local`]: super::super::Runner::add_local
+    /// [`GeneratedProcessor::descriptor`]: crate::core::processors::GeneratedProcessor::descriptor
+    #[error(
+        "add_local::<{type_name}>() failed: the type carries no processor \
+         descriptor. Annotate it with `#[processor(...)]` so it has a \
+         registerable identity."
+    )]
+    SessionProcessorHasNoDescriptor { type_name: String },
+
+    /// [`Runner::add_local`] derived a session package name that fails the
+    /// ident grammar (`[a-z][a-z0-9-]*`). `detail` carries the underlying
+    /// [`streamlib_idents::IdentError`].
+    ///
+    /// [`Runner::add_local`]: super::super::Runner::add_local
+    #[error(
+        "add_local::<{type_name}>() failed: cannot mint a `@session/<name>` \
+         ident from the type name: {detail}"
+    )]
+    SessionProcessorNameInvalid { type_name: String, detail: String },
+
+    /// [`Runner::add_local`] was handed a config that does not deserialize
+    /// into the processor type's `Config` — refused before registering, so a
+    /// session type never registers with a config its own schema rejects.
+    ///
+    /// [`Runner::add_local`]: super::super::Runner::add_local
+    #[error(
+        "add_local::<{type_name}>() failed: the supplied config is not valid \
+         for the processor's Config type: {detail}"
+    )]
+    SessionProcessorConfigInvalid { type_name: String, detail: String },
+
+    /// [`Runner::add_local`] found a live `@session/<name>` registration
+    /// already in the ledger — the same session-local name is registered and
+    /// was not removed. Never silently overwritten. Remove it first
+    /// ([`Runner::remove_module`]), or register the new type under a distinct
+    /// name.
+    ///
+    /// [`Runner::add_local`]: super::super::Runner::add_local
+    /// [`Runner::remove_module`]: super::super::Runner::remove_module
+    #[error(
+        "add_local failed: a session-local processor is already registered as \
+         '{module}'. Remove it (remove_module) before re-registering the same \
+         name, or use a distinct type name."
+    )]
+    DuplicateSessionProcessorName {
+        module: streamlib_idents::ModuleIdent,
+    },
 }
 
 impl From<AddModuleError> for Error {
--- a/runtime/streamlib-engine/src/core/runtime/module_loader/mod.rs
+++ b/runtime/streamlib-engine/src/core/runtime/module_loader/mod.rs
@@ -39,11 +39,15 @@
 //!   process-lifetime plugin-image retention.
 //! - [`ledger`] — process-global record of committed loads consumed by
 //!   [`Runner::remove_module`].
+//! - [`session`] — [`Runner::add_local`]: register a `#[processor]` host
+//!   type live under `@session/<name>@0.0.N` through the same staging seam,
+//!   with no package on disk.
 //! - [`slpkg`] — `.slpkg` archive extraction.
 //!
 //! [`Runner::remove_module`]: super::Runner::remove_module
 //! [`Runner::add_module`]: super::Runner::add_module
 //! [`Runner::add_module_with`]: super::Runner::add_module_with
+//! [`Runner::add_local`]: super::Runner::add_local
 //! [`Runner::await_modules`]: super::Runner::await_modules
 
 use std::collections::HashSet;
@@ -65,12 +69,16 @@ mod locked;
 mod processor_registration;
 mod recursive_walker;
 mod schema_registration;
+mod session;
 mod slpkg;
 mod source;
 mod staging;
 
 #[cfg(test)]
 mod tests;
+
+#[cfg(test)]
+mod session_tests;
 
 pub use added_module::{AddedModule, LoadedModule, ModuleLoadEvent};
 pub use build_orchestrator::{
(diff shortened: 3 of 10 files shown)


tato123:
Verify → PASS (no blockers). Documenting the should-fix/info items the verifier asked to record, all accepted with rationale:

**1. Engine `add_local` takes `serde_json::Value`, not the plugin-sdk `Bag` (accepted — engine purity).** `Bag` lives in the engine-free `sdk/streamlib-plugin-sdk`, which the engine cannot depend on without inverting crate layering (it would pull the cdylib-marshalling stack into the engine). `serde_json::Value` is the engine's existing config currency (`ProcessorSpec.config`), and the `Config` trait already mandates a JSON round-trip. The `Bag`-typed ergonomic surface belongs to #1412's `App` sugar (`sdk/streamlib-sdk`), which bridges `Bag → json` into this op. Owner accepted the layering rationale.

**2. Config is validate-only at registration (intentional).** `prepare_session_registration` deserializes `config` into `P::Config` purely to fail-loud on a mismatched config, then drops it — the registered `ProcessorDescriptor` carries no config; a processor instance's real config binds later at `add_processor` time (the returned `AddedModule` is a **module** handle, not an instance handle). Confirm-with-#1412 that this validate-only contract is what the App sugar expects.

**3. Test coverage — registration against a *running* runtime is a rig-gated follow-up.** The tests exercise registration on a constructed `Runner::new()` (not started). The mechanism (`spawn_blocking` + `PROCESSOR_REGISTRY`/ledger under `MODULE_REGISTRY_COMMIT_LOCK`) is runtime-state-independent, so this is a coverage gap, not a demonstrated defect — but exercising `add_local` after `Runner::start()` needs a full runtime (GPU/IPC), which the sandbox can't observe (exit 144). Filed for `/verify-live`.

**Minor (accepted, no change):** `unreachable!()` on a guarded arm and `expect()` on compile-constant invariants (`SESSION_ORG`, `"local"`) are the file's canonical idioms; `kebab_case` is a single-call-site converter (below the >1-site extraction bar, no kebab util exists yet). `warn_on_short_name_shadow`'s returned Vec is test/diagnostic-only — candidate for `pub(crate)` on a later touch.

**ABI:** no crossing — `add_local` reuses the exact `register_via_vtable(descriptor, vtable, cdylib_resident=false)` host-address-space path that in-process `register::<P>()` uses; `streamlib-plugin-abi` untouched, layout regression tests green.

PR #1459, opened 2026-07-19
feat(pack): derive processor manifests from code with a drift gate (closes #1411)

tato123:
Closes #1411

## Summary

- `sdk/streamlib-processor-extract`: cross-language (Rust/Python/Deno) processor derivation from code, plus per-platform module reachability (`reachable.rs`) so only processors reachable for the host target are compared.
- `tools/streamlib-pack`: `pkg build` now flips to the code-derived processor manifest as the source of truth, gated by `enforce_processor_manifest_matches_code` — any drift between the committed `processors:` block and what's actually reachable in code is a hard build error (`only_in_code`, `only_in_manifest`, execution/port mismatches all covered).
- `docs/decisions/manifest-extraction-capability.md` records the drift-gate + drift-surface rationale.

This is the truth-flip stage of the #1411 manifest-extraction capability (stage 1 landed in #1439, per-platform reachability landed in #1445, Python/Deno decorator inversion landed in #1440/#1448). With this PR, the manifest is no longer hand-authored truth — it's a derived, drift-checked projection of the code.

## Test evidence

```
cargo test --lib -p streamlib-processor-extract -p streamlib-pack
```

- `streamlib-processor-extract`: 59 passed, 0 failed (grammar, derive, reachable module coverage — cfg combinators, feature-gated modules, cross-platform arm resolution, parked `cfg(any())` exclusion, drift detection for only-in-code / only-in-manifest / execution-mismatch / added-port cases)
- `streamlib-pack`: 54 passed, 0 failed, including `slpkg_build_fails_on_processor_manifest_drift` (negative gate test) and `slpkg_build_succeeds_when_manifest_matches_code` (positive path)

Verified end-to-end against real in-tree packages: `packages/audio` (7 processors, platform shims present) and `packages/camera` (2 processors) both `streamlib pkg build` cleanly on Linux.

These surfaced during lens review and don't block merge, but are worth the owner's eyes:

1. **[should-fix]** `sdk/streamlib-processor-extract/src/derive.rs:464` — A target-gated (e.g. Apple-only) processor listed in a cross-platform committed manifest will false-drift on a host that doesn't compile that arm. `filter_committed_to_languages` filters committed processors by `PackageLanguage` (rust/python/deno), NOT by cfg/target. On a Linux `pkg build`, if the crate has any rust processor, Rust is in `derived_languages`, so a macOS-only rust processor in `processors:` survives the filter, is absent from the host-target derived set (reachable scan uses `for_host()`), and lands in `only_in_manifest` → hard drift error. The `enforce_processor_manifest_matches_code` doc comment (`tools/streamlib-pack/src/lib.rs`) claims "a cross-platform arm the host doesn't compile never counts as drift" — true only for the `only_in_code` direction, not `only_in_manifest`. Currently latent: no in-tree package has an Apple-only processor (Apple is parked under `cfg(any())`); verified `packages/audio` (7 procs, platform shims present) and `packages/camera` (2 procs) both `streamlib pkg build` cleanly on Linux. Suggested next step: track for Apple activation — exclude committed processors that are target-gated for a non-host triple from the `only_in_manifest` comparison, or derive across all supported targets.

2. **[should-fix]** `sdk/streamlib-processor-extract/src/derive.rs:426` — The Python and Deno extractors are spawned unconditionally on every derivation, even for packages that don't host that language. The array literal `[(Python, extractor.extract_python(package_dir)), (TypeScript, extractor.extract_deno(package_dir))]` eagerly evaluates BOTH subprocess calls when the array is constructed — the `if !languages.contains(&language) { continue; }` guard at line 430 runs only afterward, so it discards an already-spawned process rather than preventing the spawn. For the common Rust-only package this shells out to `python3 -m streamlib.extract_processors` and (if configured) `deno run ...` over a directory with no Python/Deno, throwing the output away. Correctness is preserved (skipped-language results are dropped), but every distributable `pkg build` pays needless subprocess I/O and can trigger surprising extractor invocations in environments where those runtimes are configured. Suggested next step: make the extractor call lazy under the presence guard — iterate over just the language tags, `continue` when not present, then match on `language` to invoke the corresponding `extractor.extract_*` only when actually present.

3. **[low]** `sdk/streamlib-processor-extract/src/derive.rs:82` — A package with a `Cargo.toml` but no `src/lib.rs` or `src/main.rs` crate root is not detected as Rust, so its committed rust processors are silently dropped from the drift check rather than flagged. `detect_package_languages` requires `Cargo.toml` AND (`src/lib.rs` || `src/main.rs`). A crate using a custom `[lib]` path, or a manifest listing rust processors with no crate root, passes the gate vacuously. The `slpkg_rejects_path_cargo_dependency` test relies on exactly this shape to avoid tripping the gate. Suggested next step: consider detecting the crate root from `Cargo.toml`'s `[lib]` path, or treat a `Cargo.toml` with committed rust-runtime processors but no scannable root as an error. Non-blocking.

4. **[low]** `sdk/streamlib-processor-extract/src/derive.rs:592` — `describe_difference` recomputes each `port_names(...)` `Vec<String>` twice: once in the `if` condition and again inside the `format!` body (for both inputs and outputs). Each call allocates a fresh `Vec` plus a `String` per port. Suggested next step: bind `port_names(&manifest.inputs)` / `port_names(&code.inputs)` once and reuse in both the comparison and the format (same for outputs). Cold path, purely a cleanliness nit.

5. **[info]** `sdk/streamlib-processor-extract/src/derive.rs:131` — Port schema comparison collapses to the `Type` short-name, so an org/package swap on a port schema is not detected as drift. `PortSchemaSurface::from_spec` maps both `Named(name)` and `Specific(ident)` to `Type(<short type name>)`, dropping org/package. Two schemas `@a/x/VideoFrame` and `@b/y/VideoFrame` compare equal. This is documented (module doc) and inherent — the committed bare `schema: VideoFrame` (Named) form carries no org/package — so it is the correct common denominator, but it is a known blind spot. No action suggested; recorded for awareness.

6. **[info]** `tools/streamlib-pack/src/lib.rs:530` — Negative/gate tests are non-vacuous and the gate is correctly scoped. `slpkg_build_fails_on_processor_manifest_drift` asserts the error message contains `"Camera"` AND `"source of truth"` (a string only `ManifestDriftReport::Display` emits) — reverting `enforce_processor_manifest_matches_code` makes assemble succeed and the `.unwrap_err()` panics. The call sits inside the `matches!(target, AssembleTarget::Slpkg(_))` block, so `StagedDir` (orchestrator load-time) stays exempt as documented. Derive-level tests (`only_in_code` / `only_in_manifest` / execution / added-port) each construct real drift and assert the specific report field. No action needed.

7. **[info]** `sdk/streamlib-processor-extract/src/derive.rs:464` — `filter_committed_to_languages` clones every retained `ProcessorSchema` (`.cloned().collect()`), and the sole caller (`check_processor_manifest_drift`) immediately re-projects each one to a `ProcessorSurface` and discards the owned schemas. The full-schema clone is redundant work. Optional: return `Vec<&ProcessorSchema>` (borrowing), or project straight to `Vec<ProcessorSurface>` here, so the drift check consumes the filtered surface without an intermediate owned-schema copy. Build-time path, not load-bearing — noting only.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5276,6 +5276,7 @@ dependencies = [
  "serde_yaml",
  "streamlib-cargo-build",
  "streamlib-idents",
+ "streamlib-processor-extract",
  "streamlib-processor-schema",
  "tempfile",
  "thiserror 2.0.18",
@@ -5315,7 +5316,9 @@ version = "0.7.10"
 dependencies = [
  "proc-macro2",
  "quote",
+ "serde",
  "serde_json",
+ "serde_yaml",
  "streamlib-processor-schema",
  "syn",
  "thiserror 2.0.18",
--- a/docs/decisions/manifest-extraction-capability.md
+++ b/docs/decisions/manifest-extraction-capability.md
@@ -46,22 +46,49 @@ per-language extractor imports every top-level module beside the
 decorator no longer validates its short name against a hand-authored
 `processors:` list — the decorator *is* that list.
 
-The extractors are designed to run in a fresh subprocess (`python -m
+The extractors run in a fresh subprocess (`python -m
 streamlib.extract_processors <dir>` / `deno run --allow-read
 extract_processors.ts <dir>`) so the registry starts empty; the in-process
-entrypoints clear the registry themselves. Once the pkg-build truth-flip lands,
-`pkg build` will invoke them on that path; today they ship as the standalone
-capability. Output is sorted by joined schema-ident string for determinism
-regardless of import order.
+entrypoints clear the registry themselves. Output is sorted by joined
+schema-ident string for determinism regardless of import order.
 
 The import-runs-code property is the cost of the inversion: a processor module
 whose third-party imports are unavailable cannot be enumerated, whereas the
 Rust AST scan is inert. Extraction therefore assumes the package's dependencies
-are installed — true at `pkg build` time, unlike the Rust `src/` walk. Full
-per-processor manifest fields (execution mode, entrypoint, ports) still come
-from `streamlib.yaml` for Python/Deno until ports-in-code lands for those
-runtimes; today's extractor supplies the processor identity set and declared
-port schemas, which is what the decorator carries.
+are installed — true at `pkg build` time, unlike the Rust `src/` walk.
+
+## The pkg-build drift gate
+
+`streamlib pkg build` (and `pkg publish`) derives the processor set from code —
+Rust in-process, Python/Deno via the subprocess extractors — and refuses to
+build a distributable `.slpkg` whose committed `processors:` section disagrees.
+The comparison is a **language-uniform identity surface**: processor `Type`
+name, execution mode, and each port's name + schema-type (or `any`). It
+deliberately excludes fields no code scan produces uniformly or that are
+authored/build-derived rather than code-derived — `version` (the release-core
+projection is a build concern), `entrypoint` (author/loader concern), the
+`config` binding, `description`, and the consumer-side port policies
+(`read_mode` / `overflow` / `buffer_size`, which the Python/Deno wire shape does
+not carry). What remains is exactly the surface a stale hand-authored
+`processors:` would misstate: a processor added, removed, or renamed in code; a
+port added, removed, reordered, or re-typed. The committed manifest stays the
+carrier of the excluded authored/build fields — the `processors:` section is not
+auto-overwritten, it is validated against code.
+
+Rust drift is always enforced (the `syn` scan is always runnable in-process). A
+Python/Deno extractor that cannot **run** — the runtime is absent, its import
+failed, or (Deno) it is unconfigured — is a logged skip, not a build break:
+extraction-is-import needs the runtime present, and a `pkg build` that merely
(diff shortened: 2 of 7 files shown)


tato123:
### Tracked follow-up (verifier should-fix, latent — not blocking)

`filter_committed_to_languages` / the drift comparison filters committed processors by **language** (rust/python/deno), not by **cfg/target**. A target-gated processor (e.g. an Apple-only Rust processor) listed in a cross-platform committed manifest would land in `only_in_manifest` on a Linux `pkg build` — where the host doesn't compile that arm — and hard-drift. The `enforce_processor_manifest_matches_code` doc comment's claim that "a cross-platform arm the host doesn't compile never counts as drift" holds only for the `only_in_code` direction, not `only_in_manifest`.

Currently **latent**: no in-tree package has an Apple-only processor (Apple is parked under `cfg(any())`), and camera/audio both `pkg build` cleanly on Linux. **Track for Apple activation:** exclude committed processors target-gated for a non-host triple from the `only_in_manifest` comparison, or derive across all supported targets. Filing-worthy when Apple activation lands, not before.

PR #1465, opened 2026-07-19
feat(cli): derive + reconcile package dependencies; streamlib add records the range

tato123:
Closes #1417

## Summary
- Adds a build-time dependency reconciler (`tools/streamlib-pack/src/dependency_reconcile.rs`): the referenced `@org/package` set is derived from the manifest's `schemas:` External imports plus resolved `Specific` port idents — the language-uniform, drift-validated source the catalog already uses for bare-name → owner resolution.
- Undeclared reference → hard error at package build (`Slpkg`) with a `streamlib add @org/name@<version>` fix-it.
- Declared-but-unreferenced, non-`runtime` dependency → non-fatal prune warning; kept non-destructive so a shipped manifest stays byte-identical to a `StagedDir` build.
- New `runtime: bool` marker on `DependencySpec` (Registry/Path/Git) preserves a runtime-composition dependency that imports no schema types.
- `streamlib add` is now context-sensitive on the anchor directory: in a package-authoring dir (`streamlib.yaml` with a `package:` block), `streamlib add @org/name@<version>` records a caret dependency (`^<version>`) into the package's own `dependencies:`, preserving every other manifest field and the leading `yaml-language-server` comment header. In a consumer/app dir it keeps the existing `streamlib_modules/` byte-source adoption flow.
- Regenerates the manifest JSON schema (`schemas/streamlib.schema.json`) for the new `runtime` marker.
- Docs: `docs/decisions/dependency-reconciliation.md` records the manifest-driven derivation + non-destructive prune + `runtime: true` rationale; `docs/architecture/package-development-model.md` documents the context-sensitive `add` verb.

## Test evidence
- `cargo test -p streamlib-pack --lib`: 61 passed, 0 failed (includes `slpkg_build_fails_on_undeclared_schema_dependency`, `slpkg_build_prunes_are_non_fatal_and_runtime_marker_is_honored`, `slpkg_build_succeeds_when_manifest_matches_code`, `dependency_reconcile::tests::unreferenced_dep_is_pruned_unless_runtime`).
- `cargo test -p streamlib-cli --bins`: 43 passed, 0 failed (includes `commands::add::tests::record_dependency_range_writes_caret_and_preserves_fields`).
- Verify pass confirmed a revert-fail on the undeclared-dependency hard error (removing the manifest dep reintroduces the build failure), locking the regression.

1. **should-fix** — `tools/streamlib-cli/src/commands/add.rs:163` — `record_dependency_range` preserves the manifest on `streamlib add`.
   Evidence: it reads the manifest, splits off only the leading comment header, then serde-reserializes `StreamlibYaml` and re-prepends the header. `serde_yaml` emits struct fields in `StreamlibYaml` declaration order and drops all non-leading comments, so a first `streamlib add` on a hand-authored manifest can reorder every top-level key and delete inline comments. The test only asserts the leading `# yaml-language-server` line and field *presence* survive (`record_dependency_range_writes_caret_and_preserves_fields`), not ordering or inline comments. The decision doc scopes preservation to "every other manifest field and the leading comment header" — a documented tradeoff, not a correctness bug, but it produces noisy real-world diffs.
   Suggested next step: ship as a documented PR review item; consider a comment-preserving YAML edit (e.g. `serde_yaml::Value` patch or a targeted dependencies-block insert) if authoring-diff noise matters to the owner.

2. **info** — `tools/streamlib-cli/src/commands/add.rs:32` — The new `eprintln!` for the ignored `--expect-sha256` warning violates the "no eprintln!" non-negotiable.
   Evidence: `clippy -p streamlib-cli --bins` produced zero disallowed-macros violations despite the crate's heavy `println!` use (`install.rs`/`schema.rs`/`setup.rs`); the CLI binary is de-facto exempt as user-facing output. This is the first non-test `eprintln!` in the CLI but is consistent with the existing `println!` convention and introduces no CI failure.
   Suggested next step: no action required; flagged for owner awareness.

3. **info** — `tools/streamlib-pack/src/dependency_reconcile.rs:58` — Deriving the referenced set from the manifest (not a code scan) fully covers code refs.
   Evidence: `derive_referenced_packages` unions `schemas: External{package}` imports with processor `PortSchemaSpec::Specific` idents; `PortSchemaSpec::Named` (bare names) resolve to a package only via the `schemas` map, and `Any` carries no package — both correctly handled. A bare code ref to an external schema is unusable without a `schemas`-map entry, so the manifest is the complete resolution point. Rationale is captured in `docs/decisions/dependency-reconciliation.md`.
   Suggested next step: none; noting the derivation-source design choice for the domain lens to sanity-check against the manifest-extraction pipeline (#1345/M34).

4. **low** — `tools/streamlib-cli/src/commands/add.rs:169` — Two avoidable clones (`pkg_ref.clone()` at :169, `range.clone()` at :171) exist only because both values are moved into `dependencies.insert(...)` yet reused in the `println!("{verb} dependency {pkg_ref} {range}")` / `Manifest:` lines afterward.
   Evidence: `insert()` consumes `pkg_ref` and `range`, then :183-184 print them again, forcing the clones.
   Suggested next step: cold path (single manifest write) so this is a nit. If touched: compute the print strings before the insert, or `insert(pkg_ref.clone(), ...)` only for the key and borrow the range from a local before moving — or simply print after insert using references into `manifest.dependencies`. Not worth a dedicated change.

5. **low** — `tools/streamlib-pack/src/lib.rs:822` — `enforce_declared_dependencies_match_code` iterates `reconciliation.undeclared` twice — once to build `fix_its` (:822) and once to build `refs` (:828) — allocating two `Vec`s and two joined strings over the same small slice.
   Evidence: two consecutive `.iter().map(...).collect::<Vec<_>>().join(...)` chains over the identical `undeclared` set.
   Suggested next step: trivial (error path, tiny N). If consolidated, a single loop or a `fold` could build both strings in one pass, but the current form is clearer; leave unless the block grows.

6. **info** — `tools/streamlib-pack/src/dependency_reconcile.rs:48` — `derive_referenced_packages` / `reconcile_package_dependencies` are clean, deterministic (`BTreeSet`/`BTreeMap` ordering), and reuse existing core types (`PackageRef`, `SchemaEntry::External`, `PortSchemaSpec::Specific`, `Manifest::package_ref`) rather than spinning up a parallel abstraction — consistent with engine doctrine.
   Evidence: combinator-based set derivation with self-package removal; no manual index loops, no `unwrap` in library code (unwraps confined to `#[cfg(test)]`).
   Suggested next step: none — recorded as a positive coverage observation.

### DIFF
--- a/docs/architecture/package-development-model.md
+++ b/docs/architecture/package-development-model.md
@@ -383,6 +383,16 @@ instead of a copy — recorded as a `LockfileSource::Link` — so checkout edits
 are live on the next run; `streamlib install` re-creates that symlink when
 reproducing the lock.
 
+The CLI `streamlib add` verb is context-sensitive on its anchor directory. The
+byte-source adoption above is the **consumer / app** flow (no `package:` block
+in the anchor's `streamlib.yaml`). In a **package-authoring** dir (a
+`streamlib.yaml` *with* a `package:` block), `streamlib add @org/name@<version>`
+instead records a caret dependency (`^<version>`) into that package's own
+`dependencies:` table — the schema-tier `cargo add` — preserving every other
+manifest field and the leading `# yaml-language-server` comment header. The
+`AppModulesDir` primitive itself is unchanged: it is the consumer flow and still
+refuses a registry-coordinate byte source.
+
 At load time, `Strategy::InstalledCache` (the bare `Runner::add_module`
 default) probes `<cwd>/streamlib_modules/@org/name` **before** the
 installed-package cache; an active engine `streamlib link --engine` still
--- a/docs/decisions/dependency-reconciliation.md
+++ b/docs/decisions/dependency-reconciliation.md
@@ -1,0 +1,61 @@
+# Dependency reconciliation at package build
+
+## Trigger
+
+Reach for this when touching how a package's declared `dependencies:` relates to
+what its code references — deriving the dependency set, deciding what an
+undeclared or unreferenced dependency does at `pkg build`, or wondering why the
+reconciler reads the manifest rather than re-scanning code.
+
+## Decision
+
+`pkg build` reconciles the hand-declared `dependencies:` against the dependency
+set *derived* from the package's references, for the distributable `.slpkg`
+target only (`streamlib_pack::dependency_reconcile`):
+
+- A **referenced-but-undeclared** package is a hard error, carrying a
+  `streamlib add @org/name@<version>` fix-it. The manifest must declare every
+  package its code references.
+- A **declared-but-unreferenced** package that is not marked `runtime: true` is
+  a non-fatal prune warning — dead-weight in the dependency list.
+- `runtime: true` on a dependency keeps a runtime-composition dependency that
+  imports none of the referenced package's schema types.
+
+The referenced set is derived from the **manifest** — every `schemas:
+External { package }` import plus any resolved `Specific(SchemaIdent)` port id —
+not from a fresh per-language code scan. The committed `processors:` block is
+already pinned to code by the processor-manifest drift gate, so deriving from
+the manifest *is* deriving from code, and the manifest's `schemas:` map is the
+only place a bare `schema: VideoFrame` reference's owning `@org/package` is
+recorded.
+
+## Rejected alternatives
+
+- **Scan code for the referenced set (the processor-extract crate).** The
+  cross-language processor surface collapses each port schema to its bare `Type`
+  short-name and drops `@org/package`, so it cannot name the owning dependency;
+  a Rust-only raw scan could, but the owning package for a bare port name lives
+  in the manifest's `schemas:` map regardless. Deriving from the manifest is
+  both language-uniform and the only complete source.
+- **Prune destructively — rewrite the shipped manifest to drop the dead
+  dependency.** A published `.slpkg` manifest must stay byte-identical to the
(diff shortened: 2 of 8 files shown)


tato123:
### Verify PASS — three notes for owner awareness

1. **`streamlib add` rewrites the whole manifest (should-fix, UX).** `record_dependency_range` reads the manifest, serde-reserializes `StreamlibYaml`, and re-prepends only the leading comment header — so a first `streamlib add` on a hand-authored `streamlib.yaml` can **reorder every top-level key and drop inline comments**. It's a documented tradeoff (see `docs/decisions/dependency-reconciliation.md`), correct but diff-noisy. A format-preserving edit (targeted dependencies-block insert, or a `serde_yaml::Value` patch) would keep the file tidy — I can do that as a quick follow-up (or before merge if you'd rather).

2. **`add`-verb collision resolved autonomously.** The subagent couldn't reach you mid-run, so it took the plan's recommendation: `streamlib add` **context-detects on the `package:` block** to distinguish authoring-mode (record a dep range) from consumer-mode (add a package). Flagging for confirm.

3. **First `eprintln!` in the CLI** (the ignored `--expect-sha256` warning). CI-clean — the CLI bins are exempt from the disallowed-macros lint (user-facing output), consistent with the crate's existing `println!` use. No action needed; noting since it's a first.

PR #1480, opened 2026-07-20
feat(engine): channel-centric iceoryx2 transport, zero-copy fan-out

tato123:
Closes #1419

## Summary

Inverts the intra-node iceoryx2 transport from destination-centric per-link services to
channel-centric services keyed on the **source output port** (`{source}/{source_output_port}`,
grammar landed by the #1419 D1 precursor, PR #1478). One source output port now maps to exactly
one iceoryx2 data service: a single publisher fans one zero-copy loan out to its N
compile-time-known subscribers (one per `connect()` link) plus one reserved tap slot
(`RESERVED_TAP_SUBSCRIBER_SLOTS_PER_CHANNEL`). The paired Event (notify) service stays
destination-keyed so a destination still waits on one listener fd regardless of fan-in.

Deleted (not gated):
- The fan-out copy loop in `OutputWriterInner::write_raw` — one frame + one loan + one send now,
  then notify all destination notifiers.
- The first-connection-wins subscriber guard — each link binds its own subscriber to its local
  input port, and receive routes by that binding, not the frame's port key.
- `MAX_FANIN_PER_DESTINATION` / `MAX_SUBSCRIBERS_PER_DESTINATION` and the 8x shared-memory
  reclaim workaround.
- The per-destination fan-in cap (`reject_overcap_destination_fanin`) and the shared-ring
  `DoesNotSupportRequestedMinBufferSize` startup-crash class (channels now size ring depth from
  their single source schema, not a shared destination-max).

`open_or_create_service` takes `max_subscribers` (= destination fan-out + reserved tap) with
`max_publishers` pinned to `MAX_PUBLISHERS_PER_CHANNEL` (1); `open_or_create_notify_service` takes
`max_notifiers` (= destination fan-in). Per-channel overflow is agreed across the channel's
destinations, surfacing a named `Error::Configuration` on conflict.

The Python-native and Deno-native cdylibs mirror the inversion in lockstep (one publisher + N
notifiers, N port-bound subscribers, source-port frame stamping, threaded
`max_subscribers` / `notify_max_notifiers`) so iceoryx2's open-time attribute verification stays
consistent cross-language. Inputs carry `channel_service_name` + `max_subscribers` and bind each
subscriber to its local port; outputs carry `channel_service_name` + `max_subscribers`; both carry
`notify_max_notifiers` (destination fan-in) so the subprocess opens the destination-keyed notify
service with the attributes the host used.

No plugin-ABI bump — `#[repr(C)]` / `FrameHeader` wire layout is unchanged (204 bytes); this is a
topology + service-open-attribute change, not a wire-format change. Layout regression tests pass
unchanged.

Design deviation from the plan of record (sound, matches the Zenoh key-expression research the
owner reviewed pre-approval): multi-subscriber receive routes by the subscriber's local-port
binding, not by `FrameHeader.port_key` — `port_key` now stamps the **source** port as provenance
(previously the destination port). Port-key demux is structurally impossible for source-centric
fan-out (one channel can feed N different destination ports). The live receive path
(`receive_pending`) uses the binding; `InputMailboxesInner::route()` still routes by the header's
stamped port and is used only by manual-injection callers (SDK e2e, unit tests) that frame with
the local port directly — it is off the iceoryx2 receive path, so the inconsistency is latent, not
a live bug (see review items below).

Net diff (engine + SDK commits): 16 files, +1144/-1498 (~1400-line net reduction from the deleted
copy loop and destination-centric machinery).

## Sequencing

Builds on the two merged precursors from the same milestone (Channel-Native Transport, M36):
- #1478 — D1 grammar precursor (`source_channel_name`, `/`-separated `ChannelName`), inert w.r.t.
  transport.
- #1477 — #1430's connect-time + per-frame `SchemaIdentWire` agreement check, preserved and
  re-threaded onto the re-keyed channel service in this PR (not dropped).

## Test evidence

- Full workspace unit/integration test suite green in-sandbox, including the LIVE iceoryx2 shared-
  memory revert-locks that exercise real iceoryx2 service-open attribute enforcement (a 2nd
  publisher on the same channel fails to open; the `(N + tap + 1)`th subscriber fails to open):
  - `runtime/streamlib-engine/src/iceoryx2/node.rs::channel_service_single_publisher_n_plus_tap_subscribers`
  - `runtime/streamlib-engine/src/core/compiler/compiler_ops/open_iceoryx2_service_op.rs::channel_service_name_is_source_port_shaped`
  - `.../open_iceoryx2_service_op.rs::channel_max_subscribers_counts_destinations_plus_tap`
  - `.../open_iceoryx2_service_op.rs::destination_fanin_counts_inbound_links`
  - `runtime/streamlib-engine/src/iceoryx2/output.rs::write_raw_calls_notifier`
- `#[repr(C)]` layout regression tests (`FrameHeader`, plugin-ABI vtables) pass unchanged —
  confirms no wire-format break despite the topology inversion.
- Python-native and Deno-native cdylib wiring updated in the same PR and exercised by the existing
  subprocess wiring unit tests; both sides thread `max_subscribers` / `notify_max_notifiers`
  identically to the Rust host so iceoryx2's open-time pub/sub-count verification stays consistent
  cross-language.

## E2E / live-rig status

GPU + real iceoryx2 shared-memory runtime and the Python/Deno subprocess FFI paths need a live rig
run — a sandboxed session cannot observe GPU/IPC at runtime (exit 144). `/verify-live` is flagged
as a follow-up for the owner to run on the rig before merge.

The independent verify pass (all lenses) returned PASS — mergeable. The items below are review
findings for the owner to see in-context; none block the PASS verdict.

1. **[should-fix]** `runtime/streamlib-engine/src/iceoryx2/output.rs:29` — Rename sweep is
   incomplete. The module doc still carries the intra-doc link
   `` [`OutputWriterInner::add_connection`] `` but that method was renamed to
   `set_channel_publisher`/`add_channel_notifier` (`grep 'fn add_connection'` returns nothing).
   `input.rs:24` still references the removed `set_subscriber` (now `add_channel_subscriber`).
   Broken intra-doc link violates the warning-free rustdoc rule. Same stale names also linger in
   untouched files: `runtime/streamlib-plugin-abi/src/vtables/input_mailboxes.rs:37`,
   `vtables/output_writer.rs:77`, `core/processors/processor_instance_factory.rs:456` and `:473`.
   Suggested next step: sweep the doc comments to the new method names
   (`set_channel_publisher` / `add_channel_notifier` / `add_channel_subscriber`) and fix the
   broken intra-doc link so `cargo doc` stays warning-free.

2. **[should-fix]** `runtime/streamlib-engine/src/iceoryx2/output.rs:183` — The ticket acceptance
   test "1→N fan-out to all subscribers zero-copy (loan-count assertion); N→1 fan-in via N
   channels" isn't fully delivered by name. No test exercises actual 1→N delivery:
   `write_raw_calls_notifier` (`output.rs:433`) wires exactly ONE subscriber + ONE notifier, so
   mentally reverting `write_raw`'s single-loan fan-out back to a per-destination copy loop leaves
   it green — it locks nothing about the headline single-loan-to-N guarantee.
   `channel_service_single_publisher_n_plus_tap_subscribers` (`node.rs`) only asserts subscriber
   *capacity*, and `destination_fanin_counts_inbound_links` only counts links; neither delivers a
   frame to N subscribers nor fans N channels into one input port. The startup-order-permutation
   and Python/Deno subprocess-path acceptance tests are also absent (subprocess spawn is
   reasonably `/verify-live`-deferred; the pure-Rust 1→N and N→1 delivery tests are CI-runnable
   and missing). Suggested next step: add a Rust test that publishes ONE frame through
   `OutputWriterInner::write_raw` into a service with N subscribers and asserts all N receive it
   (loan-count / single-send), plus an `InputMailboxesInner` test binding two channel subscribers
   to one local port and asserting both deliver into the shared mailbox.

3. **[should-fix]** `runtime/streamlib-engine/src/core/compiler/compiler_ops/open_iceoryx2_service_op.rs:287`
   — `channel_max_subscribers` and `channel_enable_safe_overflow` duplicate the identical
   "destinations of this source channel" graph traversal with the same filter predicate.
   `channel_max_subscribers` (l.287-289) does
   `` v(source_proc_id).out_e().iter().filter(|link| link.from_port().port_name == source_port).count() ``;
   `channel_enable_safe_overflow` (l.318-320) runs the exact same
   `` .out_e().iter().filter(|link| link.from_port().port_name == source_port) `` then
   `.map(...).collect()`. The predicate `link.from_port().port_name == source_port` IS the
   definition of a channel's destination set — if channel-membership semantics change, both sites
   must change together. Suggested next step: extract
   `fn channel_destinations(graph, source_proc_id, source_port) -> Vec<(ProcessorUniqueId, String)>`.
   `channel_max_subscribers` becomes `channel_destinations(...).len() + RESERVED_TAP_...`;
   `channel_enable_safe_overflow` iterates the returned vec. One definition of "this channel's
   destinations".

4. **[low]** `runtime/streamlib-engine/src/core/compiler/compiler_ops/open_iceoryx2_service_op.rs:308`
   — The new conflicting-overflow `Error::Configuration` validation is production-grade in shape
   but untested: `channel_enable_safe_overflow` returns a named `Error::Configuration` when a
   source port's destinations declare conflicting overflow policies, but no unit test exercises
   that error path. Suggested next step: add a graph test wiring one source port to two
   destinations with block vs. drop_oldest input-port overflow and assert `Error::Configuration`.

5. **[low]** `runtime/streamlib-engine/src/iceoryx2/output.rs:93` — `has_channel_publisher` and
   `has_port` now have byte-identical bodies since both structs collapsed onto the single
   `channels` map. l.93-94 `has_channel_publisher` returns
   `self.channels.lock().contains_key(output_port)`; l.179-180 `has_port` returns
   `self.channels.lock().contains_key(port)`. After the connections→channels rename a
   port-with-a-publisher and a configured-port are the same predicate, so these two public methods
   can silently diverge in meaning while sharing an implementation. Suggested next step: have
   `has_channel_publisher` delegate to `has_port` (or drop one), keeping a single source of truth
   for "this output port is configured".

6. **[info]** `runtime/streamlib-engine/src/iceoryx2/input.rs:464` — Frame-header port routing is
   not fully retired in favor of binding-based routing. The live receive path
   (`receive_pending`) now routes by the subscriber's local-port binding, and `write_raw` stamps
   the SOURCE port into the header (previously the dest port) — a semantic inversion of the
   `port_key` field with the byte layout unchanged (layout regression tests pass unchanged,
   satisfying the "no wire-header change" criterion). `route()` (`input.rs:464`) still routes by
   the header's stamped port and is used only by manual-injection callers (SDK e2e, unit tests)
   that frame with the local port; it is off the iceoryx2 receive path, so the inconsistency is
   latent, not a live bug. Suggested next step: optionally reconcile or document `route()` as
   manual-injection-only.

7. **[info]** `runtime/streamlib-engine/src/core/compiler/compiler_ops/open_iceoryx2_service_op.rs:536`
   — Wiring-envelope helpers have a clean signature but a large one:
   `wire_subprocess_source`/`wire_subprocess_dest` each take 9 positional args (4 of them sizing
   numbers) behind `#[allow(clippy::too_many_arguments)]`; a small sizing-params struct would read
   better. Internal-only, both SDK sides consistent. Suggested next step: optional — group the
   sizing params into a named struct.

8. **[info]** `runtime/streamlib-engine/src/core/compiler/compiler_ops/open_iceoryx2_service_op.rs:564`
   — The subprocess-host downcast dispatch (Deno else-if Python) is repeated across
   `is_subprocess_processor`, `wire_subprocess_source`, and `wire_subprocess_dest` — a
   shotgun-surgery point when a third subprocess runtime is added. `wire_subprocess_source`
   (l.564-571) and `wire_subprocess_dest` (l.606-613) each `downcast_mut::<DenoSubprocessHostProcessor>()`
   then `else if downcast_mut::<PythonNativeSubprocessHostProcessor>()` to push onto
   `output_port_wiring` / `input_port_wiring`; `is_subprocess_processor` (l.404) downcasts again. A
   new SDK runtime forces edits in every such site. Suggested next step: optional (predates this
   PR, and this PR already reduced the inline duplication by building the envelope once) — reify a
   `SubprocessHostProcessor` trait with `push_output_wiring(&mut self, entry)` /
   `push_input_wiring(&mut self, entry)` so the dispatch collapses to one downcast returning
   `&mut dyn SubprocessHostProcessor`. Not blocking.

## Follow-ups

- #1421, #1420 unblock once this merges (both were `blocked_by` #1419).
- #1430 stays open: criterion 3 (strict schema-agreement opt-in) is a small follow-up once this
  PR's per-channel surface is on `main`.


tato123:
### Post-verify hardening applied (verify was PASS; these are the enumerated should-fixes, one fix round, no re-verify)

- **1→N fan-out + N→1 fan-in DELIVERY tests** — `write_raw_fans_out_single_loan_to_all_subscribers` (3 subscribers each receive exactly one frame from a single loan+send) is **empirically revert-locked**: reverting `write_raw` to a per-connection copy loop makes it fail (a subscriber gets 2 frames). `two_channel_subscribers_fan_into_one_local_port` locks N→1. Both run against live iceoryx2 shm.
- **Fan-out microbench** (`fanout_1_to_n`, N∈{1,2,4,8}) — per-delivered-frame throughput stays flat/improves (~345K→388K elem/s), the single-loan signature vs the retired copy loop.
- **Warning-free rustdoc** — fixed the broken `add_connection` intra-doc link + swept stale method names (`set_subscriber`→`add_channel_subscriber`, etc.) across the touched files + plugin-abi vtables + processor factory.
- **`channel_destinations()`** extracted as the single definition of a channel’s destination set (DRY; `channel_max_subscribers` + `channel_enable_safe_overflow` both call it); conflicting-overflow `Error::Configuration` now has a test.

### Reviewer notes
- **`port_key` semantic flip (byte layout unchanged):** the header now stamps the **source** port as provenance; the live receive path (`receive_pending`) routes by the subscriber→local-port binding, not `port_key` — a single publisher can’t stamp a per-destination key. `FrameHeader` is 204 B unchanged, so the layout regression tests + the “no wire-header change” criterion still hold. `route()` is retained for the manual-injection path (SDK e2e + unit tests) with its doc corrected.
- **Optional, not done (noted):** the 9-positional-arg subprocess wiring helpers could take a sizing-params struct; the Deno/Python downcast dispatch could become a `SubprocessHostProcessor` trait. Both internal-only, non-blocking.
- **⚠️ `/verify-live` required before merge:** the pure-Rust 1→N/N→1 delivery + startup-order locks pass in-sandbox, but full N-way fan-out delivery and **Python/Deno subprocess round-trips** need a rig run (sandbox can’t observe the GPU/IPC pipeline at runtime).

PR #1482, opened 2026-07-20
feat(engine): dynamic iceoryx2 PowerOfTwo slot allocation; retire authored max_payload_bytes

tato123:
Closes #1421

## Summary

Retires authored `max_payload_bytes` in favor of dynamic PowerOfTwo iceoryx2 slot allocation.
Publishers now start from a small size hint and grow their loan size to the next power of two
on demand; subscribers grow-and-retry their read buffer instead of failing when a payload
exceeds the current slot capacity. This deletes the crash class where an encoder's first IDR
frame (or any payload above the old fixed `MAX_PAYLOAD_SIZE`) was rejected with
`ExceedsMaxLoanSize`.

- Engine (`runtime/streamlib-engine/src/iceoryx2/{input,output,node}.rs`): dynamic PowerOfTwo
  publisher sizing + grow-and-retry subscriber reads; per-channel trust-tier ceiling
  (`ChannelTrustTier` + `TRUSTED_CHANNEL_PAYLOAD_CEILING_BYTES`) refuses runaway growth instead
  of allocating unbounded memory.
- `streamlib-ipc-types`: fixed-size `#[repr(C)] FramePayload` deleted — the transport is
  slice-based; `MAX_PAYLOAD_SIZE` renamed to `DEFAULT_EXPECTED_PAYLOAD_BYTES`. New
  `decide_channel_egress_admission` is the single authority for per-channel ceiling refusal +
  PowerOfTwo growth bookkeeping, called from the host `OutputWriterInner::write_raw` and both
  subprocess natives (`slpn`, `sldn`) instead of three independently-maintained copies.
- Plugin ABI: `INPUT_MAILBOXES_VTABLE_LAYOUT_VERSION` bumped 2 → 3 (new vtable slot for
  grow-and-retry reads; `OutputWriter` unchanged); layout regression + version-pin tests updated
  (size 48 → 40, `max_payload_for_port` slot removed); ABI fingerprint auto-rotates. The
  grow-and-retry read loop itself is shared via `streamlib_plugin_abi::grow_and_retry_read`
  between the engine's `InputMailboxes::read_raw` and the SDK's, retiring the earlier duplicate.
- Python/Deno native FFI (`streamlib-python-native`, `streamlib-deno-native`): PowerOfTwo
  publishers + grow-and-retry reads ported to both subprocess-native runtimes, keeping lockstep
  with the engine-side protocol.
- `streamlib-error`: new `Error::PayloadExceedsChannelCeiling` typed variant (replaces the old
  unbounded-allocation failure mode with a named, catchable error); `tier` is now a typed
  `ChannelTrustTierLabel` enum (engine-free, `Display`-backed) instead of a stringly-typed
  `String` filled via `.as_str().to_string()`.
- `ChannelEgressConfig` struct replaces the positional channel-egress argument bag on
  `set_channel_publisher` / `wire_rust_source` (was 7–8 positional args under
  `#[allow(clippy::too_many_arguments)]`); the `#[allow]` is dropped.
- Node-level `max_payload_bytes_per_channel` deployment ceiling: per-tier env override
  (`STREAMLIB_MAX_PAYLOAD_BYTES_PER_CHANNEL_TRUSTED` /
  `STREAMLIB_MAX_PAYLOAD_BYTES_PER_CHANNEL_UNTRUSTED_SESSION`), read/parsed in the engine so
  `ipc-types` stays logging-free; unset is byte-identical to the tier default, a bad value warns
  and falls back (never panics).
- SDK facade: `ChannelTrustTier` + `TRUSTED_CHANNEL_PAYLOAD_CEILING_BYTES` routed through
  `streamlib-sdk` so `hello_streamlib_example_e2e` doesn't reach past the SDK boundary into raw
  `streamlib-engine` paths (`check-boundaries` gate).
- Example/package schema comments and `spawn_processor_op`'s stack-size rationale updated where
  the deleted `MAX_PAYLOAD_SIZE`/`FramePayload`/`ExceedsMaxLoanSize` crash class was documented
  as current behavior.
- Latest round: quarter-ceiling crossing folded into the egress authority so all three runtimes
  (host writer, Python native, Deno native) warn consistently; the channel-egress write tracing
  is now shared across the natives via `emit_channel_egress_admission_tracing`; a
  `ChannelTrustTier::as_str()` field replaces the previously hardcoded tracing tier literals; and
  a drift-guard test (`trust_tier_label_spellings_do_not_drift`) pins `ChannelTrustTier::as_str`
  against `ChannelTrustTierLabel`'s `Display` so the two spellings of the trust-tier labels can't
  diverge silently.

## Test evidence

- Full local gate battery green on the current branch tip, including 28/28 `iceoryx2` module
  tests and 76/76 `streamlib-plugin-abi` tests.
- Empirical probe (gating first implementation step, per doctrine — source-reading vs claim):
  live segment regrow followed by the subscriber succeeded before the rest of the change was
  built on top of it. `publisher_grows_segment_for_oversized_loan_and_delivers`
  (`runtime/streamlib-engine/src/iceoryx2/node.rs`) primes a 64-byte hint, loans 1 MiB, and
  asserts the subscriber receives the full grown payload byte-for-byte; fail-without-fix is
  non-vacuous (dropping the `AllocationStrategy::PowerOfTwo` line reverts to `Static` →
  `ExceedsMaxLoanSize`).
- Revert-locked tests added for growth, ceiling refusal, and grow-and-retry staging
  (`runtime/streamlib-engine/src/iceoryx2/{input,output,node}.rs`,
  `core/embedded_schemas/integration_tests.rs`) — fail without the fix, per
  `feedback_regression_tests_must_exercise_bug`.
- Python (`pytest`) and Deno (`deno test`) grow-and-retry FFI paths exercised in-sandbox, including
  a fake-native-driven unit test of the `SLPN_READ_NEEDS_LARGER_BUFFER` / grow-and-retry loop
  (`sdk/streamlib-python/python/streamlib/tests/test_payload_sizing.py`) and the Deno
  `context_test.ts` equivalent.
- Node-level ceiling env override covered by a defensive-parse test (unset → byte-identical
  default; bad value → warn + fallback, never panic).
- `trust_tier_label_spellings_do_not_drift` locks `ChannelTrustTier::as_str` against
  `ChannelTrustTierLabel` Display; git-grep confirmed the `max_payload_bytes` sweep is complete
  (only an unrelated webrtc RTP-MTU local constant remains).
- `verify-change` (HIGH-risk, ABI-touching) ran multiple times on this branch as it evolved —
  all lenses clear on the final pass.

No live-rig `/verify-live` run was required for this PR: the change is exercised through
in-process/in-sandbox iceoryx2 shared-memory tests (including the empirical regrow probe) and
Python/Deno subprocess round-trips, not through GPU/camera/display/codec hardware paths. The
true cross-FFI subprocess growth probe (a real Python or Deno subprocess consuming a multi-MB
keyframe over iceoryx2) is rig-gated and not exercised by this PR's in-sandbox coverage — see
review items below.

- **info** — `runtime/streamlib-engine/src/iceoryx2/node.rs:880`: PowerOfTwo dynamic slot
  allocation with transparent subscriber remap is delivered and proven against real iceoryx2.
  Evidence: `create_publisher` now sets `.allocation_strategy(AllocationStrategy::PowerOfTwo)`;
  test `publisher_grows_segment_for_oversized_loan_and_delivers` primes a 64-byte hint, loans
  1 MiB, and asserts the subscriber receives the full grown payload byte-for-byte. Ran green
  (28/28 iceoryx2 tests). Fail-without-fix is non-vacuous: dropping the strategy line reverts to
  Static -> ExceedsMaxLoanSize. Suggested next step: None — mechanism verified in-process.

- **info** — `runtime/streamlib-plugin-abi/src/vtables/input_mailboxes.rs:24`: ABI change is
  correctly versioned with updated layout regression + version-pin tests in the same PR
  (INPUT_MAILBOXES v2->v3, OutputWriter unchanged; size 48->40, max_payload_for_port slot
  removed). Evidence: layout test asserts size_of == 40 and drops the max_payload_for_port
  offset; `input_mailboxes_vtable_layout_version_pinned_at_three` passes; lib.rs layout_tests
  pins INPUT_MAILBOXES_VTABLE_LAYOUT_VERSION == 3. plugin-abi tests 76/76 green. Suggested next
  step: None.

- **info** — `runtime/streamlib-ipc-types/src/lib.rs:88`: Ceiling refusal + growth bookkeeping is
  centralized in one authority (decide_channel_egress_admission) and shared by host + Python +
  Deno natives, with a drift-guard test — matching engine "one system per concern" doctrine
  rather than three parallel copies. Evidence: host output.rs, slpn_output_write,
  sldn_output_write all call `decide_channel_egress_admission`; tracing shared via
  `emit_channel_egress_admission_tracing`; `trust_tier_label_spellings_do_not_drift` locks
  ChannelTrustTier::as_str against ChannelTrustTierLabel Display. Crash class replaced by named
  `PayloadExceedsChannelCeiling`; #1407 `BagPayloadOverflow`/`BAG_MAX_PAYLOAD_BYTES` deleted;
  `FramePayload` removed; constant renamed to DEFAULT_EXPECTED_PAYLOAD_BYTES; full
  `max_payload_bytes` sweep confirmed by git grep (only unrelated webrtc RTP-MTU local const
  remains). Suggested next step: None.

- **info** — `runtime/streamlib-engine/src/iceoryx2/output.rs:88`: current_slot_capacity_bytes
  growth tracking is a best-effort mirror of iceoryx2's internal segment size, not a source of
  truth. Evidence: growth events log old/new segment bytes computed as
  frame_total_bytes.next_power_of_two() on an engine-side tracked value primed to
  expected_payload_bytes+FRAME_HEADER_SIZE; documented as "best-effort tracking ... so a growth
  event is observable". Feeds tracing only, never admission correctness (the ceiling compares
  actual total_len). Suggested next step: None — acceptable for an observability field; owner
  may confirm the logged segment sizes are understood as approximate.

- **should-fix** — `runtime/streamlib-engine/src/iceoryx2/output.rs:227`: The host writer's
  write_raw hand-rolls the same three channel-egress tracing events that the shared plugin-abi
  helper emit_channel_egress_admission_tracing already owns — a maintained twin of logic the PR
  otherwise deliberately centralized. Evidence: output.rs L227-266 emits tracing::warn "output
  channel refused a payload above its per-channel ceiling", tracing::info "iceoryx2 publisher
  data segment grew (PowerOfTwo)", and tracing::warn "iceoryx2 publisher segment crossed a
  quarter of the channel ceiling" with field names channel/payload_bytes/ceiling_bytes/tier/
  refused_count and old_segment_bytes/new_segment_bytes/segment_bytes.
  runtime/streamlib-plugin-abi/src/vtables/output_writer.rs L106-152
  (emit_channel_egress_admission_tracing) emits byte-identical message strings and field names
  for the Python/Deno natives. The three call sites read the SAME decide_channel_egress_admission
  result but two copies of the emission code must be kept in lock-step by hand; the added
  trust_tier_label_spellings_do_not_drift test guards only the tier string, not these message
  strings — a reworded warning or an added field drifts silently between host and subprocess
  logs. Suggested next step: Give emit_channel_egress_admission_tracing a trust_tier:
  ChannelTrustTier parameter (replacing the hardcoded UntrustedSession) and an optional
  log-prefix (None for the host, Some(("slpn"/"sldn", processor_id)) for the natives), then call
  it from output.rs::write_raw and delete the inline match block — collapsing three copies of the
  emission into one authority alongside the decision it already shares.

- **info** — `runtime/streamlib-plugin-abi/src/vtables/output_writer.rs:122`:
  emit_channel_egress_admission_tracing hardcodes tier = ChannelTrustTier::UntrustedSession.as_str()
  at all three log sites rather than deriving it, which is correct today only because subprocess
  channels are definitionally untrusted-session. Evidence: Lines 122, 135, 145 all pass
  ChannelTrustTier::UntrustedSession.as_str() unconditionally. This is the concrete reason the
  host (which also handles the Trusted tier) cannot reuse the helper and instead maintains the
  twin flagged above. A subprocess is always untrusted-session so no bug exists, but the
  hardcode is a latent mislabel trap and couples the helper to one caller class. Suggested next
  step: Fold into the should-fix: thread the real tier through as a parameter so the helper is
  tier-agnostic and reusable by the host writer.

- **low** — `sdk/streamlib-deno-native/src/lib.rs:606`: The grow-and-retry peek block (next-index
  selection + required-length check + NEEDS_LARGER_BUFFER return) is duplicated verbatim between
  the Deno and Python natives. Evidence: sdk/streamlib-deno-native/src/lib.rs L606-621 and
  sdk/streamlib-python-native/src/lib.rs ~L623-641 are identical: `let next_index = if read_mode
  == READ_MODE_READ_NEXT_IN_ORDER { 0 } else { queue.len() - 1 }; let required =
  queue[next_index].0.len(); if required > buf_len as usize { ... return
  *_READ_NEEDS_LARGER_BUFFER }`. The two natives are parallel-by-design FFI clones, but this PR
  already extracts other newly-shared native logic (emit_channel_egress_admission_tracing) into
  plugin-abi, so leaving this new block un-extracted is inconsistent. The queue.len()-1 is safe
  (guarded by the preceding is_empty() continue). Suggested next step: Optional: extract a small
  `fn peek_front_or_latest_len(queue: &[(Vec<u8>, i64)], read_mode: i32) -> usize` (or return the
  index) into plugin-abi so the two natives share the peek logic the same way they share the
  admission decision; leave as-is if the natives are slated for a broader de-duplication pass.


tato123:
### Verify verdict: PASS. Applied vs deferred

verify-change PASSED. Retitled `refactor`→**`feat`** (the vtable-layout ABI bump 2→3 + `max_payload_bytes` retirement means a `refactor` title would silently skip the release version bump).

**Applied in a follow-up fix round (completing this PRs canonical sweep — cheap, zero-logic):**
- stale comments referencing the deleted `MAX_PAYLOAD_SIZE`/`FramePayload`/`ExceedsMaxLoanSize` swept in the two `vulkan-video-roundtrip*` example manifests (matching moq-roundtrip + vulkan-video-psnr already done here) and `spawn_processor_op.rs`;
- dead `max_payload_bytes` keys renamed → `expected_payload_bytes` in the tool test fixtures;
- named the `0..8` grow-and-retry bound → `MAX_GROW_AND_RETRY_ATTEMPTS`.

**Deferred to a batched follow-up (structural — per ship-on-verify-pass, not blocking; call your merge):**
- **DRY #5:** the grow-and-retry read loop is duplicated verbatim between engine `InputMailboxes::read_raw` and the plugin-sdk twin — same ABI protocol both sides, must change together. Extract one `grow_and_retry_read(...)` helper into `streamlib-plugin-abi`.
- **DRY #7:** `set_channel_publisher` (7 args) / `wire_rust_source` (8 args) carry the four channel-egress primitives (`service_name`, `trust_tier`, `expected_payload_bytes`, `ceiling_bytes`) as a positional bag behind `#[allow(too_many_arguments)]` → introduce `struct ChannelEgressConfig`.
- **#6 native twin (explicit decision):** the ceiling-refusal + PowerOfTwo-growth block is triplicated across engine `write_raw` + `deno-native` + `python-native`. The two natives already twin `PublisherState`/`READ_MODE`/`init_subprocess_logging`, so this rides the established subprocess-native duplication pattern — **flagging it as an explicit decision, not drift.** Collapsing the two native copies into a shared subprocess-native helper is the clean follow-up.

**Notes (info, no action):**
- **FramePayload deleted vs the tickets "re-scope" wording:** it was dead on the wire (slice-based transport since #1419), so deletion is cleaner than rescoping dead code. Confirm youre fine with the deletion.
- **Trust-tier ceiling is STRUCTURAL** (subprocess end → UntrustedSession 16 MiB, host↔host → Trusted 64 MiB); no explicit node-level deployment-override knob yet — the 2026-07-18 design mentioned a "node-level ceiling", so the override is a clean config follow-up. Confirm the structural-only shape is acceptable.
- `Error::PayloadExceedsChannelCeiling` carries `tier: String` (not the `ChannelTrustTier` enum) because `streamlib-error` is a leaf crate that cant depend on `streamlib-ipc-types` — acknowledged intentional.

**⚠️ `/verify-live` for #1421:** the Python/Deno FFI grow-and-retry paths were exercised in-sandbox (pytest + deno test) + live-iceoryx2-shm growth tests pass, so this is better-covered than #1419 — but a full subprocess round-trip of an oversized frame end-to-end still merits a rig run.

tato123:
### ✅ UPDATE — all should-fixes now applied on-branch (supersedes the "deferred" note above)

Per owner direction (2026-07-20): should-fix = fixed before merge, not deferred. The three structural should-fixes previously marked deferred in comment above are now **applied on this branch** (fix round, gates 17/0, no re-verify — the branch was already verify-PASS; these are the verifiers own enumerated items + gate battery confirms):

- **grow-and-retry DRY** → `streamlib_plugin_abi::grow_and_retry_read(...)` shared helper; engine host + engine-free SDK twin both call it (`MAX_GROW_AND_RETRY_ATTEMPTS` moved into the helper).
- **`ChannelEgressConfig`** → replaces the 7/8-arg positional bag on `set_channel_publisher` / `wire_rust_source`; `#[allow(clippy::too_many_arguments)]` dropped; all 5 call sites migrated.
- **native ceiling/growth triplication** → single authority `decide_channel_egress_admission()` in `streamlib-ipc-types` (logging-free); engine `write_raw` + `slpn_output_write` + `sldn_output_write` all call it; each keeps only its own runtime-tagged tracing + refusal surface (host typed `PayloadExceedsChannelCeiling` vs natives rc-2), now an intentional per-boundary split, not accidental drift.

**No should-fix remains open on this PR.** The two `info` items still stand as owner-confirm-at-merge notes (FramePayload deleted vs "re-scope" wording; structural-only trust-tier with no node-level override knob yet — a config-knob follow-up). Ready to merge.

tato123:
### ✅ Validation converged — verify PASS, all should-fixes fixed (4 validation passes)

Per owner direction (should-fix = fix-before-merge + validate changes with independent review, not edit-and-park), #1482 went through four verify-change passes; each surfaced *different* real items and the change was re-validated after every should-fix. Final verify **PASS**, all should-fixes cleared. Disposition:

**Fixed on-branch (all gate-validated; final gates 35/0):**
- **node-level per-channel ceiling deployment override** (the 2026-07-18 design’s "node-level ceiling") — env-var override on the tier default, defensive parse (invalid → warn + fall back, no panic), byte-identical when unset.
- **typed the ceiling-error tier** (`ChannelTrustTierLabel` enum) — dropped the `String` + alloc.
- **quarter-ceiling growth warn parity** — was host-only; folded the threshold into the shared `decide_channel_egress_admission` so **all three runtimes** (engine + Python + Deno natives) warn.
- **one tracing authority** — the refusal/growth/quarter-ceiling tracing is now a single tier-agnostic `emit_channel_egress_admission_tracing` in `plugin-abi`, used by host + both natives (was triplicated).
- **native `tier` label via `ChannelTrustTier::as_str()`** (drift-guarded) instead of hardcoded literals.
- shared `grow_and_retry_read` + `next_read_required_len` peek across host + natives; `ChannelEgressConfig` struct; drift-guard + Deno grow-retry tests; earlier stale-comment/fixture sweep.

**Left, with reason:**
- **⚠️ `/verify-live` (rig):** the polyglot subprocess growth path is proven **structurally in-sandbox** (fake-native grow-and-retry tests, live-iceoryx2-shm host growth) but NOT via a real Python/Deno subprocess consuming a multi-MB keyframe over iceoryx2 — that E2E is rig-gated. Recommend a `/verify-live` run.
- **best-effort growth-segment tracking** and the `decide_...` `&mut`-counter shape — disclosed/justified (info), no change.

**One owner call at merge:** the node-override **shape** — I implemented per-tier env vars (`STREAMLIB_MAX_PAYLOAD_BYTES_PER_CHANNEL_TRUSTED` / `_UNTRUSTED_SESSION`) matching the codebase’s deployment-tunable pattern; a typed node-config struct is a single-swap-point alternative if you prefer. No should-fix remains. Ready to merge.

PR #1486, opened 2026-07-20
feat(sdk): unify read_mode/overflow/buffer_size/max_queued_messages into one delivery profile

tato123:
Closes #1420

## Summary

Unifies the four smeared delivery knobs (`read_mode`, overflow policy, buffer sizing,
`max_queued_messages`) into a single per-port `DeliveryProfile`: `latest`
(SkipToLatest + DropOldest, depth 4), `every_sample` (ReadNextInOrder + DropOldest,
counted+warned drops, depth 16), `lossless` (ReadNextInOrder + Block, depth 16).

- New `DeliveryProfile` core system (`runtime/streamlib-engine/src/iceoryx2/delivery_profile.rs`) —
  the one place profile → `(read_mode, overflow, depth)` resolution lives.
- Defaults derived from a per-vocabulary-type `flow_class` carried in the schema `metadata` block
  (`sample_stream` → `every_sample`, `state_stream` → `latest`; `lossless` is explicit-only),
  replacing `read_mode`/`max_queued_messages` there.
- One-word code override at the port declaration site — Rust `#[processor]` attribute + Python/Deno
  decorators (parity tests added for both).
- The four legacy knobs and every dead/hardcoded `read_mode` path deleted across Rust engine,
  plugin-sdk, Python-native, Deno (canonical-pattern sweep in this PR); every schema YAML's
  `metadata` swept to `flow_class`.
- `schemas/streamlib.schema.json` regenerated for the new manifest `delivery_profile` port field;
  manifest→descriptor path threads it through (`30e56840`, `3b81e8ae`).
- No plugin-ABI version bump — the wire representation is `Option<String>` end to end, not a new
  `#[repr(C)]` field.

71 files changed, +793/-692.

## Test evidence

Ran in a clean worktree at the branch head (`3b81e8ae`):

- CI's own unit-test gate (`.github/workflows/test.yml`):
  `cargo test --locked -p streamlib -p streamlib-plugin-abi -p streamlib-plugin-sdk -p streamlib-processor-extract --lib`
  → **all green**: `streamlib` 2/2, `streamlib-plugin-abi` 77/77, `streamlib-plugin-sdk` 101/101 (1 ignored),
  `streamlib-processor-extract` 60/60 (includes the new `grammar::tests::input_delivery_profile_is_accepted` /
  `output_delivery_profile_is_rejected` cases).
- `cargo test -p streamlib-engine --lib delivery_profile::` → **7/7 passed** (per-profile resolution table,
  `flow_class` default-vs-landmine mapping, manifest string round-trip, unknown-profile/lossless-as-flow-class
  rejection).
- `cargo test -p streamlib-engine --lib embedded_schemas::` → **31/31 passed** (`delivery_profile_defaults_from_flow_class`,
  `delivery_profile_override_wins_over_flow_class`, `delivery_profile_defaults_latest_when_processor_unregistered`,
  `delivery_profile_rejects_unknown_override`, plus the pre-existing registry/payload-bytes suite unaffected by the sweep).
- `cargo run -p xtask -- check-manifest-schema` → **57/57 `streamlib.yaml` files validated** against the regenerated
  `schemas/streamlib.schema.json`.

## E2E / live-verify report

Rig-touching (touches the iceoryx2 transport read/overflow/depth path), so a live rig run was done on this
milestone's merged transport stack (#1419/#1421/#1430) immediately ahead of this ticket's implementation to
confirm the underlying channel-native transport delivers frames end-to-end before layering the profile sweep
on top: built `camera-display` at `0.7.19`, ran `vivid(/dev/video0) → tatolab/camera → iceoryx2 transport →
tatolab/display`, screenshotted the live display window — vivid SMPTE color bars rendered correctly, OSD
timecode advancing, 1920x1080 NV12, transport log clean (no schema-mismatch, no ceiling-refusal/growth error,
no panic), graceful shutdown. Evidence: https://gh-artifact.tatolab.com/share/20260720-194407/m36-channel-transport-vivid-smpte.jpg

A dedicated live run of *this* branch's per-profile behavior under a throttled consumer (does `latest` skip,
does `every_sample` bound+drop-oldest, does `lossless` backpressure) was not additionally captured — see
should-fix item 2 below; the resolution logic itself is unit-tested end to end.

- **[should-fix]** `packages/opus/src/opus_decoder.rs:189` (same pattern at `webrtc_whip.rs:40`) — the sweep
  silently changes encoded-audio delivery semantics: `encoded_audio_in` previously declared
  `read_mode="skip_to_latest"` (drop stale packets, keep newest) and now falls back to the `EncodedAudioFrame`
  flow_class default `every_sample` (`ReadNextInOrder`, in-order). This diverges from how the same PR faithfully
  preserved `read_next_in_order`-on-`VideoFrame` by adding an explicit `delivery_profile="every_sample"` override
  (h264/h265 encoder `video_in`, `video_frame_counter`) — the `skip_to_latest` divergences were dropped instead of
  mapped to `delivery_profile="latest"`. Evidence: `git diff e94ef4c0..branch` — `opus_decoder` & `webrtc_whip`
  `encoded_audio_in` lose `read_mode = "skip_to_latest", buffer_size = 8` and gain no override; `EncodedAudioFrame`
  schema now `flow_class: sample_stream` → `every_sample`. No test asserts the resulting delivery for either
  encoded-audio port. Suggested next step: decide intent — if the old skip-to-latest was a deliberate live-stream
  latency bound, restore it with `delivery_profile="latest"` on both `encoded_audio_in` ports; if the
  `every_sample` canonicalization is intended, state that explicitly (encoded audio now drains in-order rather
  than skipping to latest).
- **[should-fix]** `runtime/streamlib-engine/src/iceoryx2/delivery_profile.rs:146` — the acceptance criterion
  "Tests: per-profile semantics under a slow consumer (latest skips; every_sample bounds+drops-oldest; lossless
  backpressures)" is not covered; only the resolution table, `flow_class` default resolution, override-wins, and
  the conflict error are unit-tested — nothing exercises actual iceoryx2 delivery behavior under a slow consumer.
  Runtime IPC can't run in the sandbox (exit 144), so this is genuinely rig-gated. Suggested next step: a
  live-verify pass (or an ignored integration test) exercising latest-skips / every_sample-drops-oldest /
  lossless-backpressures against a throttled reader on the rig; the resolution logic itself is well covered so
  this is a documented deferral, not a blocker.
- **[info]** `sdk/streamlib-macros/src/codegen.rs:706` — the `#[processor]` macro no longer registers Rust
  in-process input ports (dropped the `__post_install add_port` calls with `buffer_size`/`read_mode`); the host
  wire path (`wire_rust_dest`) now owns all port registration via `add_port(dest_port, depth, drain_order)` using
  the resolved delivery profile. This correctly removes the dead/hardcoded `read_mode` path the ticket called out.
  Only verifiable at runtime — folds into the live-verify note above; confirm during live verification that
  in-process Rust processors still receive frames on all wired inputs.
- **[should-fix]** `runtime/streamlib-engine/src/core/embedded_schemas/mod.rs:210` — the newly-added
  `flow_class_for_port_spec` copy-pastes the entire registry-lookup preamble of the pre-existing
  `resolve_metadata_u64_for_port_spec` (`mod.rs:325`): the `Any`→`Ok(None)` guard, `schema_spec.to_string()`,
  the `get_embedded_schema_definition(...).ok_or_else(...)` miss handling, and the `serde_yaml::from_str` parse.
  The multi-line "referenced by a port spec but not in the runtime schema registry — did you forget to call
  `runtime.add_module(...)`" error string is duplicated verbatim (`mod.rs:221` and `mod.rs:335`). Same-logic-must-
  change-together duplication: reword the add_module hint and you must edit two sites. Suggested next step:
  extract a `metadata_value_for_port_spec(schema_spec, field) -> Result<Option<&serde_yaml::Value>>` helper that
  owns the `Any`-guard + registry-miss error + yaml parse, with both resolvers projecting only their leaf; call
  this out as replacing real duplication (per the no-silent-DRY-refactor rule).
- **[low]** `runtime/streamlib-engine/src/iceoryx2/delivery_profile.rs:22` — `DeliveryProfile` (and `FlowClass`,
  ~line 96) derive `Serialize, Deserialize` with `#[serde(rename_all = "snake_case")]`, but neither is ever
  serialized/deserialized via serde: the value crosses every wire boundary as `Option<String>`
  (`PortInfo`/`PortDescriptor`/`ProcessorPortSchema`) and is parsed exclusively through the hand-rolled
  `from_manifest_str`. The serde mapping is a silent second source of truth for the same strings that
  `as_manifest_str`/`from_manifest_str` already own. Suggested next step: drop the unused derives, or route
  `from_manifest_str` through serde if round-tripping is actually intended.
- **[low]** `runtime/streamlib-engine/src/core/embedded_schemas/mod.rs:228` — `flow_class_for_port_spec` swallows
  a YAML parse failure with `Err(_) => return Ok(None)`, silently falling back to the `Latest` default. This
  contradicts the same function's own rustdoc ("a wire-time error, not a silent default substitution") and its
  loud registry-miss handling three lines above. Embedded schemas are pre-validated so it rarely fires, but a
  malformed `metadata` block would resolve a channel to the wrong delivery profile with no diagnostic. Suggested
  next step: map the parse error to `Error::Configuration` naming the schema, consistent with the registry-miss
  branch — do it in the extracted helper from the should-fix above.

### DIFF
--- a/docs/architecture/static-registry.md
+++ b/docs/architecture/static-registry.md
@@ -159,12 +159,9 @@ version (a missing dep is `ExternalDepMissing`, a cycle is
 
 ### Deliberate omissions
 
-`CatalogPort` carries `name`, `description`, `schema`, `read_mode` only. Two
-classes of manifest field are intentionally dropped:
+`CatalogPort` carries `name`, `description`, `schema`, `delivery_profile` only.
+One class of manifest field is intentionally dropped:
 
-- `overflow` / `buffer_size` — present on the authored manifest port, but
-  they are per-edge *buffering knobs*, not wiring topology, so they don't
-  belong in a palette the editor wires graphs from.
 - `required` — has no authored-manifest source at all; it exists only on the
   runtime port descriptors, not on the `streamlib.yaml` port.
 
--- a/docs/decisions/manifest-extraction-capability.md
+++ b/docs/decisions/manifest-extraction-capability.md
@@ -67,9 +67,9 @@ name, execution mode, and each port's name + schema-type (or `any`). It
 deliberately excludes fields no code scan produces uniformly or that are
 authored/build-derived rather than code-derived — `version` (the release-core
 projection is a build concern), `entrypoint` (author/loader concern), the
-`config` binding, `description`, and the consumer-side port policies
-(`read_mode` / `overflow` / `buffer_size`, which the Python/Deno wire shape does
-not carry). What remains is exactly the surface a stale hand-authored
+`config` binding, `description`, and the consumer-side `delivery_profile`
+(which the Python/Deno wire shape does not carry uniformly). What remains is
+exactly the surface a stale hand-authored
 `processors:` would misstate: a processor added, removed, or renamed in code; a
 port added, removed, reordered, or re-typed. The committed manifest stays the
 carrier of the excluded authored/build fields — the `processors:` section is not
--- a/docs/learnings/startup-crash-iceoryx2-wire-vs-gpu-setup-race.md
+++ b/docs/learnings/startup-crash-iceoryx2-wire-vs-gpu-setup-race.md
@@ -72,7 +72,8 @@ doesn't fire.
 ## Reference
 
 - iceoryx2 sizing fix: `core/compiler/compiler_ops/open_iceoryx2_service_op.rs`
-  (`max_queued_messages_for_dest`) + `iceoryx2/node.rs` (`max_subscribers`).
+  (the channel's agreed delivery-profile ring depth) + `iceoryx2/node.rs`
+  (`max_subscribers`).
 - The latent GPU race's mechanism (main-vs-fan-out glcore contention) and the
   funnel candidate fix live in the tracked issue for it, not here — this file
   is the *diagnostic* learning, not the fix proposal.
--- a/packages/audio/src/apple/audio_output.rs
+++ b/packages/audio/src/apple/audio_output.rs
@@ -55,7 +55,7 @@ pub struct AppleAudioDevice {
     execution = manual,
     scheduling = realtime,
     config = crate::_generated_::AudioOutputConfig,
-    input("audio", "@tatolab/core/AudioFrame", read_mode = "read_next_in_order", buffer_size = 32, description = "Stereo audio frame to play through speakers"),
+    input("audio", "@tatolab/core/AudioFrame", description = "Stereo audio frame to play through speakers"),
 )]
 pub struct AppleAudioOutputProcessor {
     device_id: Option<usize>,
--- a/packages/audio/src/audio_channel_converter.rs
+++ b/packages/audio/src/audio_channel_converter.rs
@@ -12,7 +12,7 @@ use streamlib_plugin_sdk::sdk::context::{RuntimeContextFullAccess, RuntimeContex
     execution = reactive,
     scheduling = realtime,
     config = crate::_generated_::AudioChannelConverterConfig,
-    input("audio_in", "@tatolab/core/AudioFrame", read_mode = "read_next_in_order", buffer_size = 32, description = "Mono audio frame"),
+    input("audio_in", "@tatolab/core/AudioFrame", description = "Mono audio frame"),
     output("audio_out", "@tatolab/core/AudioFrame", description = "Multi-channel audio frame"),
 )]
 pub struct AudioChannelConverterProcessor {
--- a/packages/audio/src/audio_mixer.rs
+++ b/packages/audio/src/audio_mixer.rs
@@ -12,8 +12,8 @@ use streamlib_plugin_sdk::sdk::context::{RuntimeContextFullAccess, RuntimeContex
     execution = reactive,
     scheduling = realtime,
     config = crate::_generated_::AudioMixerConfig,
-    input("left", "@tatolab/core/AudioFrame", read_mode = "read_next_in_order", buffer_size = 32, description = "Left channel mono audio frame"),
-    input("right", "@tatolab/core/AudioFrame", read_mode = "read_next_in_order", buffer_size = 32, description = "Right channel mono audio frame"),
+    input("left", "@tatolab/core/AudioFrame", description = "Left channel mono audio frame"),
+    input("right", "@tatolab/core/AudioFrame", description = "Right channel mono audio frame"),
     output("audio", "@tatolab/core/AudioFrame", description = "Mixed stereo audio frame"),
 )]
 pub struct AudioMixerProcessor {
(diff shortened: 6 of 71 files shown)


tato123:
### Verify PASS → fixed + LIVE-verified. Ready to merge (closes #1420, completes M36).

**Fixed on-branch (verifier findings 2/3/4; gates green):**
- `metadata_value_for_port_spec` shared helper — collapsed the duplicated registry-lookup preamble that `flow_class_for_port_spec` had copy-pasted from the u64 resolver (incl. the verbatim `add_module` error string).
- Malformed schema-YAML now → named `Error::Configuration` (was a silent `Latest` fallback contradicting the fn’s own rustdoc); locked by a regression test.
- Dropped the unused `Serialize/Deserialize` derives on `DeliveryProfile`/`FlowClass` (never serialized via serde; `from_manifest_str`/`as_manifest_str` is the single mapping).

**Finding 1 (encoded-audio "skip_to_latest regression") — FALSE POSITIVE, grounded and closed:** the change-verifier read `EncodedAudioFrame` as `sample_stream`, but `packages/core/schemas/encoded_audio_frame.yaml` declares `flow_class: state_stream` → `FlowClass::StateStream => DeliveryProfile::Latest` → `Latest.resolve()` = `ReadMode::SkipToLatest`. So dropping the explicit `read_mode="skip_to_latest"` on the opus/webrtc `encoded_audio_in` ports **preserves** skip-to-latest via the type default; adding an override would violate the sweep’s "pin only deviations" rule. Cross-checked every pre-sweep `read_mode` against its wire type’s flow_class default — every genuine deviation was already correctly pinned (h264/h265→every_sample, mp4→lossless, video_frame_counter→every_sample). No port silently lost a divergence.

**✅ LIVE rig verification PASS** (loop-run via the bypass, against this branch): built `camera-display` @ the branch, ran vivid `/dev/video0` → `tatolab/camera` → the delivery path → `tatolab/display`. vivid SMPTE bars render correctly, live OSD timecode advancing, **zero errors/panics**, graceful shutdown — the four-knob→`DeliveryProfile` sweep carries frames end-to-end and the in-process Rust display input still registers + receives frames (covers the port-registration finding). Evidence: https://gh-artifact.tatolab.com/share/20260720-221900/m1420-delivery-profile-vivid-smpte.jpg

**Deferred (rig-gated follow-up, not a blocker):** the throttled-reader profile-behavior matrix (latest-skips-to-newest / every_sample-drops-oldest-counted / lossless-backpressures under a slow consumer). The resolution logic (profile→triple, flow_class default, override-wins, conflict error) is unit-tested and the end-to-end delivery is live-confirmed; isolating the slow-consumer drop/block behavior needs a purpose-built throttled pipeline — a follow-up eval.

**Accepted design points:** `DeliveryProfile` is a NEW per-port core system (the only delivery knob), ring depths derive from the profile (per-type depths shrink where the profile consolidates them; explicit overrides preserve authors’ deviating intent), wildcard/Any ports default to `latest`/depth 4.

PR #1493, opened 2026-07-21
feat(engine): first-class channel tap op — read-only reserved-slot subscriber

tato123:
Closes #1426

## Summary

Adds `RuntimeOperations::tap_async(channel, count)` — a first-class, read-only subscriber attach on any named channel, using the `+1` reserved tap slot the channel data service is already sized for (`RESERVED_TAP_SUBSCRIBER_SLOTS_PER_CHANNEL`). No new iceoryx2 service, no publisher re-open, no sizing change. This is the "perceive" half of perceive-act (#1426).

**Engine**
- Extracted the channel-sizing derivation (`resolve_channel_sizing` + `find_channel_source_port`) shared by the service-open compiler op and the new tap op, so the tap's publisher-free reopen requests identical, iceoryx2-verified parameters (`max_subscribers` = destinations + reserved tap).
- `Iceoryx2Service::create_tap_subscriber` maps iceoryx2's `ExceedsMaxSupportedSubscribers` to a distinct `ReservedSlotOccupied` outcome, surfaced as the named `Error::TapSlotOccupied`.
- Because the iceoryx2 `Subscriber` is `!Send`, a tap owns a dedicated OS thread holding the subscriber and forwarding raw bag bytes over a tokio mpsc (the same sync-source → mpsc → async bridge the event WebSocket already uses). Detach = drop the `TapSubscription` → thread joins → reserved slot frees. The forwarder uses `try_send` (never blocking) so a stalled downstream drops bags instead of back-pressuring the source's `send()`; drop closes the receiver before `join()`.
- `RuntimeOpsShim` (subprocess/cdylib side) rejects tap with `Error::NotSupported` — there is no `RuntimeOpsVTable` op; tap is host-side by construction.
- New named error taxonomy: `Error::TapChannelNotFound` (no iceoryx2 data service wired) and `Error::TapSlotOccupied` (reserved slot already taken by a concurrent tap).

**api-server**
- `GET /ws/tap/{channel}?count=` streams raw bags as binary WS frames, gated behind the same bearer token as the mutating routes when auth is enabled.
- Detach runs via `tokio::task::spawn_blocking` so the `TapSubscription` drop (which joins an OS thread) never blocks a tokio worker.

**Scope**
- Not a `RuntimeOpsVTable` change — api-server is in-process, no cdylib hop.
- Polyglot parity is language-specific by construction: tap is host-side, so Python/Deno channels are tappable transparently with zero SDK change.
- No wire / `#[repr(C)]` / JTD change.

## Commits

- `c0fdc5e3` feat(error): named tap error taxonomy — `TapChannelNotFound` + `TapSlotOccupied`
- `a71ad6fc` feat(engine): first-class channel tap op — read-only reserved-slot subscriber
- `b37d9de9` fix(engine): tap forwarder must never back-pressure production
- `66ceca59` fix(api-server): tap WS auth-gate test, non-blocking detach, clear query param

## Test evidence

- Tap a live channel mid-run, observe bags in order; detach frees the slot (a second tap succeeds only after detach); double-tap → `TapSlotOccupied`; bounded tap delivers exactly `count` then ends; `resolve_channel_sizing` sizing-recovery + channel reverse-resolve unit locks.
- `stalled_downstream_never_blocks_the_drain_and_detach_returns_promptly` (lossless, `enable_safe_overflow=false` channel): asserts `dropped_bags() > 0` and that `drop(tap)` returns within 5s. Ran: PASS. Mentally reverting to `blocking_send` makes the dropped-bags loop hit its 10s deadline and hangs `drop(tap)`'s `recv_timeout` — non-vacuous.
- `tap_ws_rejects_missing_token_with_401_when_auth_on`, `tap_ws_with_token_clears_the_auth_gate`, `tap_ws_is_open_with_auth_off`: PASS. Deleting the `route_layer` flips the 401 assertion to the WS extractor's non-401 rejection — non-vacuous.
- api-server lib built and all 17 tests: PASS.

- **[info]** `runtime/streamlib-engine/src/core/runtime/tap.rs:246` — Prior-round backpressure finding is resolved: the forwarder must never park a lossless channel's drain. Forwarder now uses `forward_tx.try_send(...)` and on `TrySendError::Full` drops the bag (`dropped_bags` counter, warn-throttled) instead of blocking; `Drop` sets `stop_flag` + closes the receiver before `join()`. Regression test `stalled_downstream_never_blocks_the_drain_and_detach_returns_promptly` runs on a lossless (`enable_safe_overflow=false`) channel, asserts `dropped_bags()>0` and that `drop(tap)` returns within 5s. Ran it: PASS. Mentally reverting to `blocking_send` makes the `dropped_bags` loop hit its 10s deadline and drop(tap) hang the `recv_timeout` — non-vacuous. Suggested next step: None — resolved.
- **[info]** `packages/api-server/src/handlers.rs:95` — Prior-round auth-gate finding is resolved: the read-only tap WS is bearer-gated like the mutating routes. `tap_router` carries `.route_layer(from_fn_with_state(tap_auth_token, require_bearer_token))` when auth is on. Three new tests (`tap_ws_rejects_missing_token_with_401_when_auth_on`, `..._with_token_clears_the_auth_gate`, `..._is_open_with_auth_off`) ran PASS. Deleting the `route_layer` flips the 401 assertion to the WS extractor's non-401 rejection — non-vacuous. Suggested next step: None — resolved.
- **[info]** `packages/api-server/src/handlers.rs:586` — Prior-round non-blocking-detach finding is resolved: `TapSubscription::drop` (which joins an OS thread) no longer runs on a tokio worker. Detach is now `tokio::task::spawn_blocking(move || drop(subscription)).await`. api-server lib built and all 17 tests PASS. Suggested next step: None — resolved.
- **[low]** `runtime/streamlib-engine/src/core/runtime/tap.rs:243` — On sustained downstream overload the tap drops the NEWEST bag, so a slow WS/MCP client trails live indefinitely rather than skipping to fresh data, and `dropped_bags()` is never surfaced to the WS client. `tap.rs:246` drops `_dropped_bag` (the just-received newest sample) on `TrySendError::Full`; older bags stay queued in the mpsc. `handle_tap_websocket` never forwards `subscription.dropped_bags()` to the client. This is deliberate and documented (module doc lines 19-26) as a non-interference tradeoff; acceptable for the bounded-count observe use-case but suboptimal for a live observability tap under load. Suggested next step: Optional follow-up: consider drop-oldest for the live (`count=None`) case and/or emit a dropped-bags notice frame; fine to ship as a documented review note, not blocking.
- **[info]** `runtime/streamlib-engine/src/core/compiler/compiler_ops/open_iceoryx2_service_op.rs:356` — The `ChannelSizing`/`resolve_channel_sizing` extraction is a justified shared derivation, not a silent DRY refactor, and is in scope. Both the service-open compiler op (line 157) and the tap op (`operations_runtime.rs:434`) consume `resolve_channel_sizing`, and correctness requires identical sizing on the tap's publisher-free reopen (iceoryx2 verifies `max_subscribers`). Called out in doc comments + commit messages, and locked by `resolve_channel_sizing_recovers_service_open_max_subscribers` (asserts agreement with `channel_max_subscribers`). This is the engine-doctrine "extend the existing derivation" move, not a parallel abstraction. Suggested next step: None.

### DIFF
--- a/packages/api-server/src/handlers.rs
+++ b/packages/api-server/src/handlers.rs
@@ -58,10 +58,11 @@ pub(crate) fn build_router(
     auth_token: Option<ApiServerBearerToken>,
     #[cfg(feature = "moq")] runtime_id: String,
 ) -> Router {
-    // The read-only tap WebSocket sits behind the SAME bearer gate as the
-    // mutating routes when auth is on (a tap observes live channel data), and is
-    // open like every other route when auth is off. Clone the token before it is
-    // moved into the mutating-route middleware below.
+    // The read-only tap WebSocket is gated exactly like the mutating routes WHEN
+    // auth is opted in — same bearer middleware, same route_layer binding; the
+    // default (auth off) leaves it open like every other route. This is
+    // mechanism parity, not a trust boundary the tap itself imposes. Clone the
+    // token before it is moved into the mutating-route middleware below.
     let tap_auth_token = auth_token.clone();
 
     let mut protected = OpenApiRouter::new()
@@ -562,10 +563,11 @@ async fn handle_tap_websocket(
         Ok(subscription) => subscription,
         Err(e) => {
             tracing::info!(channel = %channel, "tap attach rejected: {e}");
+            let (close_code, close_reason) = tap_error_close_frame(&e);
             let _ = sender
                 .send(Message::Close(Some(axum::extract::ws::CloseFrame {
-                    code: axum::extract::ws::close_code::ERROR,
-                    reason: e.to_string().into(),
+                    code: close_code,
+                    reason: close_reason.into(),
                 })))
                 .await;
             return;
@@ -605,6 +607,48 @@ async fn handle_tap_websocket(
     }
 
     tracing::info!(channel = %channel, "tap client detached");
+}
+
+/// Longest tap close reason RFC 6455 permits: a control frame caps its payload
+/// at 125 bytes and the 2-byte close code consumes the first two, leaving 123
+/// for the UTF-8 reason. tungstenite refuses to write an over-length close
+/// frame, so an untruncated tap error string (`NotSupported` runs ~180 bytes)
+/// would drop the client into an abnormal close with no reason at all.
+const MAX_WS_CLOSE_REASON_BYTES: usize = 123;
+
+/// Map a typed tap error to a WebSocket close code + a short, RFC-6455-legal
+/// reason (≤ [`MAX_WS_CLOSE_REASON_BYTES`], truncated on a UTF-8 char
+/// boundary). The full error is logged server-side at the call site; this
+/// surface is the machine-readable failure the client (and the #1429 MCP tool)
+/// reads off the close frame. App codes live in the 4000–4999 private range.
+fn tap_error_close_frame(error: &Error) -> (u16, String) {
+    let (code, reason) = match error {
+        Error::TapChannelNotFound(channel) => {
+            (4404, format!("tap channel not found: {channel}"))
+        }
+        Error::TapSlotOccupied(channel) => {
+            (4409, format!("tap slot already occupied: {channel}"))
+        }
+        other => (
+            axum::extract::ws::close_code::ERROR,
+            format!("tap attach failed: {other}"),
+        ),
+    };
+    (code, truncate_on_char_boundary(reason, MAX_WS_CLOSE_REASON_BYTES))
+}
+
+/// Truncate `text` to at most `max_bytes`, cutting on a UTF-8 char boundary so
+/// the result stays valid UTF-8.
+fn truncate_on_char_boundary(mut text: String, max_bytes: usize) -> String {
+    if text.len() <= max_bytes {
+        return text;
+    }
+    let mut boundary = max_bytes;
+    while boundary > 0 && !text.is_char_boundary(boundary) {
+        boundary -= 1;
+    }
+    text.truncate(boundary);
+    text
 }
 
 #[cfg(test)]
@@ -919,10 +963,10 @@ mod router_auth_gate_tests {
 
     #[tokio::test]
     async fn tap_ws_rejects_missing_token_with_401_when_auth_on() {
-        // The read-only tap is a security-relevant surface: with auth on it must
-        // be gated exactly like the mutating routes. Deleting the tap_router
-        // `.route_layer(...)` flips this from 401 to the WS extractor's own
-        // (non-401) rejection, going red here.
+        // With auth opted in, the read-only tap is gated exactly like the
+        // mutating routes — mechanism parity, not a trust boundary the tap
+        // imposes. Deleting the tap_router `.route_layer(...)` flips this from
+        // 401 to the WS extractor's own (non-401) rejection, going red here.
         assert_eq!(status_of(tap_ws_request()).await, StatusCode::UNAUTHORIZED);
     }
 
--- a/packages/api-server/src/handlers.rs
+++ b/packages/api-server/src/handlers.rs
@@ -513,9 +513,9 @@ impl EventListener for WebSocketEventForwarder {
 /// Query parameters for the tap WebSocket: an optional bounded sample count.
 #[derive(Deserialize)]
 pub(crate) struct TapQuery {
-    /// Stream exactly `n` bags then close; absent streams live until the client
-    /// disconnects.
-    n: Option<usize>,
+    /// Stream exactly `count` bags then close; absent streams live until the
+    /// client disconnects.
+    count: Option<usize>,
 }
 
 /// `GET /ws/tap/{channel}` — attach a read-only tap to `channel` and stream its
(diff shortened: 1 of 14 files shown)


tato123:
## ✅ Review-items follow-up

This PR opened **after** the fix round + a clean delta re-verify (PASS), so the items above are already in their resolved state. Recapping against the *original* verify (which found the deadlock + more) for a complete picture:

- ✅ **[was blocker] drop-time deadlock** (`tap.rs`) — **FIXED.** The forwarder no longer `blocking_send`s; it `try_send`s and drops on full, and `Drop` calls `receiver.close()` before `join()`. Regression-locked: reverting to `blocking_send` makes `drop(tap)` hang >60s.
- ✅ **[was should-fix] tap back-pressures production on a lossless channel** — **FIXED** by the same change. The forwarder now *always* drains its iceoryx2 subscriber, so the source processor is never gated by a slow tap consumer. Test on a lossless (`safe_overflow=false`) channel asserts `dropped_bags() > 0` and `drop(tap)` returns within 5s.
- ✅ **[was should-fix] `/ws/tap/{channel}` auth gate untested** — **FIXED.** Three gate tests added (401 without token when auth on, passes with token, open when auth off).
- ✅ **[was should-fix] blocking `join()` in async `Drop`** — **FIXED.** Detach moved to `tokio::task::spawn_blocking`, so the WS handler's async drop never blocks a runtime worker.
- ⏸️ **[low — DEFERRED, documented, verifier-sanctioned ship]** under sustained overload the tap drops the **newest** bag (a slow client trails live indefinitely) and `dropped_bags()` isn't surfaced to the WS client. This is a deliberate non-interference tradeoff, documented in the module. **Not fixed here** — follow-up candidate: drop-oldest for the live (`count=None`) case + a dropped-bags notice frame. Your call whether to file it.
- ℹ️ **`ChannelSizing`/`resolve_channel_sizing` extraction** — confirmed a justified shared derivation (the tap's publisher-free reopen must request the exact sizing the compiler op opened with), not a silent DRY refactor.

**Merge-ready.** Live polyglot tap wire-neutrality (tap a Python/Deno-published channel) is the one deferred-to-`/verify-live` item.

tato123:
## ⛔ Live-verify found a merge blocker — tap does not work end-to-end

Ran a real hello-world publisher → iceoryx2 channel → `/ws/tap/{channel}` WebSocket end-to-end (via a standalone `@tatolab/hello-tap` package + the real `streamlib-runtime`). The whole pipeline came up (source+sink registered version-free @0.0.0, connected, iceoryx2 channel Wired) but the **tap attach was rejected**:

```
tap attach rejected: Operation not supported: tap is a host-side operation —
a subprocess cannot initiate a channel tap across the plugin ABI ... tap from the host api-server / MCP surface
```

**Root cause (confirmed in code):** the api-server processor captures `runtime = ctx.runtime()`, which for every processor — including the FullAccess/trusted api-server — is the **`RuntimeOpsShim`**. `RuntimeOpsShim::tap_async` returns `NotSupported` by design (a subprocess can't own the host's `!Send` iceoryx2 subscriber). So the WS route's `ctx.runtime().tap_async()` always hits the shim → NotSupported. The api-server's other routes work because the shim dispatches add_processor/connect fine; only `tap_async` is refused.

**Why the verify passed anyway:** all hermetic tap tests call `Runner::tap_async` **directly** (the host handle, which works), bypassing the real `api-server → shim → tap_async` path. The feature was tested at the one seam where it works, not the seam it runs through.

**Fix (real, not a doc change):** the in-process FullAccess (trusted-tier, #1427) api-server needs a **host-direct** `tap_async` (the actual `Runner`), not the shim — the shim's blanket NotSupported is correct only for the subprocess/limited tier. Also: send a proper WS Close frame on rejection (the NotSupported reason came back as an oversized control frame, >125 bytes).

**Follow-up test discipline:** add an api-server↔tap test that drives the tap through the api-server's actual runtime handle (or a live run), not `Runner::tap_async` directly.

Marking this PR blocked pending the fix.

tato123:
## ✅ verify-live PASSED — tap streams readable frames end-to-end

Ran the full chain live (real `streamlib-runtime` + api-server + a `@tatolab/hello-tap` publisher → iceoryx2 channel → `GET /ws/tap/{channel}`) with the Shape-C fix (`ddc8c880`: host-resident `ctx.runtime()` reaches the real Runner ops instead of the shim):

```
Control plane is up.
Tap URL: ws://127.0.0.1:39599/ws/tap/ppliyj3skt4g5nvdn1enr2fbj%2Foutput?count=5
tap client attached
  frame 0: seq=2 message="hello world 2"
  frame 1: seq=3 message="hello world 3"
  frame 2: seq=4 message="hello world 4"
  frame 3: seq=5 message="hello world 5"
  frame 4: seq=6 message="hello world 6"
SUCCESS: captured 5 readable frames over the tap.
```

Before the fix, `tap client attached` never happened — the attach was rejected with `NotSupported` because the api-server got the subprocess `RuntimeOpsShim`. Now it reaches `Runner::tap_async`, attaches, and streams live bags (seq starts at 2 = the tap sees only post-attach frames, correct). The `host_ctx_runtime_reaches_real_runner_ops_not_the_shim` regression test is revert-proven, and the boot-level WS test locks a well-formed close frame. **This is the merge-blocking e2e break from earlier, resolved and proven.**

PR #1499, opened 2026-07-21
feat(packages): MCP veneer over the api-server

tato123:
Closes #1429

## Summary
- Adds an MCP (Model Context Protocol) veneer over the existing api-server: a JSON-RPC `/mcp` endpoint advertising 7 tools — `graph`, `submit_processor`, `replace_processor`, `remove_processor`, `connect`, `tap`, `logs`.
- The MCP surface reuses the api-server's existing bearer-token auth (`crate::auth::require_bearer_token`) and the same runtime the REST API drives — no new auth path, no parallel runtime handle.
- Extracts the shared submit/replace-processor-source conversion into `crate::ops` so REST and MCP both call the same composite instead of re-implementing it (real duplication removal, called out per-commit).
- Deletes the dead, already-commented-out `[[bin]] streamlib-mcp` binary config from `runtime/streamlib-engine/Cargo.toml`, superseded by this veneer.
- Adds `tap_async` to `RollbackObservingStubRuntime`, the api-server test stub — this was a pre-existing compile gap from #1493 (the trait added `tap_async` with no default body) that left the api-server test binary non-compiling on the merge base; this branch fixes it.

## Commits
- `569406d8` chore(engine): delete dead streamlib-mcp binary config
- `f198de43` refactor(api-server): extract shared submit/replace source ops
- `65943d00` feat(packages): MCP veneer over the api-server (#1429)

## Test evidence
- `cargo test -p streamlib-api-server --lib` → 32 passed / 0 failed, including:
  - `tools_list_advertises_every_veneer_tool` — asserts all 7 tools are advertised.
  - `tools_call_submit_processor_reaches_the_runtime` — asserts the recorded source text reaches `register_processor_source_async` with `module=@session/widget@=0.0.0`.
  - `mcp_endpoint_is_gated_by_bearer_auth_when_enabled` — asserts 401 with no token and 200 with a valid token (non-vacuous: removing the `route_layer` flips the 401 to 200).

These are the verified findings from the independent verify pass. None block the acceptance criteria (list tools + submit-processor end-to-end, both delivered and tested); several are flagged `should-fix` for a fast follow-up rather than this PR.

1. **[should-fix]** `packages/api-server/src/mcp.rs:409` — `call_tap` bridges a live channel stream to a request/response result with a bag-count bound but **no time bound**: a tap on an idle or slow channel blocks the MCP request indefinitely until `sample` bags arrive (bounded only by the client's request timeout). Asymmetric with `call_logs`, which is bounded by a monotonic sample window. Suggested fix: wrap `subscription.recv()` in `tokio::time::timeout_at(deadline, ...)` and return the partial sample when the window elapses.

2. **[should-fix]** `packages/api-server/src/mcp.rs:390` — the streaming→bounded-sample bridge logic in `tap` and `logs` (the most non-trivial code in the file) and the `replace_processor`/`remove_processor`/`connect` dispatchers have no MCP-layer test. `mcp::tests` only covers `initialize`, `tools/list`, `tools/call submit_processor`, `graph`, unknown-tool/method, and auth. Suggested follow-up: a tap-tool test (stub `tap_async` returning N bags, assert `byte_len`/`hex_preview`/`hex_truncated`/`dropped_bags`), a logs-tool test (publish via PUBSUB, assert the bounded sample), and thin dispatch tests for remove/connect/replace.

3. **[should-fix]** `packages/api-server/src/mcp.rs:319` — the `SubmittedProcessorSource` wire-DTO→runtime-type conversion (`source`, `language.into()`, `requested_name`, `processor_type_name`) is copy-pasted across 4 sites: `mcp.rs:319`, `mcp.rs:339`, `handlers.rs:331`, `handlers.rs:402` — exactly the duplication the `ops.rs` extraction was meant to prevent, left unswept at 2 of the 4 sites. Suggested fix: an inherent `into_submitted_source(self)` on the two request DTOs (a blanket `From` impl is orphan-blocked), called at all 4 sites.

4. **[should-fix]** `packages/api-server/src/mcp.rs:547` — `McpEventForwarder` is a byte-for-byte copy of `handlers.rs`'s `WebSocketEventForwarder` (same field, same `EventListener` impl, same pubsub-Event→unbounded-mpsc bridge); the doc comment even admits it mirrors the REST forwarder. Suggested fix: hoist one explicit forwarder type (zero-context name, e.g. `PubSubEventToChannelForwarder`) into a shared module and use it from both call sites.

5. **[low]** `packages/api-server/src/mcp.rs:185` — the processor-language enum is now hardcoded in a third location (alongside the SDK `ProcessorLanguage` schema and `ProcessorLanguageDto`'s utoipa schema in `state.rs:129`); adding a 5th language and missing one spot degrades to a client-side JSON Schema validation mismatch, not a wrong dispatch. Also: `rust` is advertised in the enum but rejected downstream for live source submit (`state.rs:110-111`), with no description note warning the caller. Optional: derive the MCP enum from the same source `state.rs` builds, and add a clause noting `rust` is wire-parity-only.

6. **[low]** `packages/api-server/src/mcp.rs:419` — the tap-detach idiom (`spawn_blocking(move || drop(subscription))`, required because `TapSubscription::drop` joins an OS thread that must not join on a tokio worker) is copied verbatim from `handlers.rs:766`, including the warn-on-join-failure branch. This non-obvious concurrency rule is now stated in two places. Optional: extract a shared `async fn detach_tap_subscription(...)`.

7. **[low]** `packages/api-server/src/mcp.rs:433` — `call_logs` takes `state: &AppState` that it never uses, suppressed with `let _ = state;` in the body (it subscribes to the global PUBSUB, not `AppState`). Optional: rename the parameter to `_state` and drop the body statement.

8. **[low]** `packages/api-server/src/mcp.rs:4` — the module docs in `mcp.rs` and `ops.rs` are multi-paragraph, brushing against the "no multi-paragraph rustdoc essays" comments rule, though the content is non-derivable wire-contract/why rationale rather than narration. Optional: tighten to 1-2 lines.

9. **[info]** `packages/api-server/src/mcp.rs:492` — `submitted_source_json` hand-maps every field of `SubmittedSourceOutcome` into a `json!` literal rather than deriving `Serialize` and calling `to_value`; defensible since the HTTP front end projects the same struct into a different response shape (adds `events_url`). No action needed.

10. **[info]** `packages/api-server/src/mcp.rs:536` — `hex_encode` is hand-rolled rather than pulling in a `hex` crate dependency for one preview helper. No action needed.

11. **[info]** `packages/api-server/src/mcp.rs:447` — `call_logs` subscribes the global `PUBSUB` per call (not `state.runtime`), spawning a short-lived OS thread each invocation. Correct in the single-runtime api-server process; the poll loop exits promptly on an idle bus once the listener's `Arc` is dropped. No action needed.

12. **[info]** `packages/api-server/src/ops.rs:234` — `session_processor_ident` builds a `SchemaIdent` from a minted `@session/name@=0.0.N` module's exact pinned version with no prerelease stripping; safe because session versions never carry a prerelease and the ident never enters the catalog/schemas registry tree. Code moved verbatim from `handlers.rs`, unchanged behavior. No action needed.

13. **[info]** `runtime/streamlib-engine/Cargo.toml:228` — the branch carries three scope-adjacent changes beyond the MCP module itself (dead-binary-config deletion, the `ops.rs` extraction, and the `tap_async` stub fix), each called out in its own commit message. The `tap_async` fix addresses a pre-existing #1493 api-server test-compile gap; owner may want to track that separately, but nothing further to change here.

### DIFF
--- a/packages/api-server/Cargo.toml
+++ b/packages/api-server/Cargo.toml
@@ -59,5 +59,9 @@ getrandom = "0.2"
 constant_time_eq = "0.3"
 
 [dev-dependencies]
+# Enables the engine's `test-support` in-memory `TapSubscription` constructor
+# for the MCP/REST tap-tool tests. A dev-dep feature: active for tests, absent
+# from the production `cargo build` dependency graph.
+streamlib = { path = "../../sdk/streamlib-sdk", version = "0.7.23", features = ["test-support"] }
 tempfile = "3"
 tower = {version = "0.5", features = ["util"]}
--- a/packages/api-server/src/handlers.rs
+++ b/packages/api-server/src/handlers.rs
@@ -11,15 +11,13 @@ use axum::{
     extract::ws::{Message, WebSocket, WebSocketUpgrade},
     http::StatusCode,
     response::IntoResponse,
-    routing::get,
+    routing::{get, post},
 };
 use serde::Deserialize;
 use futures_util::{SinkExt, StreamExt};
 use parking_lot::Mutex;
 use std::sync::Arc;
-use streamlib::sdk::descriptors::{
-    ModuleIdent, Org, Package, SchemaIdent, SemVer, SemVerRange, TypeName,
-};
+use streamlib::sdk::descriptors::{Org, Package, SchemaIdent, SemVer, TypeName};
 use streamlib::sdk::error::{Error, Result};
 use streamlib::sdk::graph::{InputLinkPortRef, OutputLinkPortRef};
 use streamlib::sdk::json_schema::{
@@ -29,10 +27,7 @@ use streamlib::sdk::json_schema::{
 use streamlib::sdk::processors::PROCESSOR_REGISTRY;
 use streamlib::sdk::processors::ProcessorSpec;
 use streamlib::sdk::pubsub::{Event, EventListener, PUBSUB, topics};
-use streamlib::sdk::runtime::{
-    RegisterProcessorReceipt, ReplaceProcessorFromSource, RuntimeOperations,
-    SubmittedProcessorSource,
-};
+use streamlib::sdk::runtime::{RuntimeOperations, SubmittedProcessorSource};
 use tower_http::trace::{DefaultMakeSpan, DefaultOnRequest, DefaultOnResponse, TraceLayer};
 use tracing::Level;
 use utoipa::OpenApi;
@@ -42,9 +37,7 @@ use crate::auth::{ApiServerBearerToken, ForbiddenResponse, UnauthorizedResponse}
 use crate::state::{
     ApiDoc, AppState, CreateConnectionRequest, CreateProcessorRequest, ErrorResponse, IdResponse,
     ProcessorNotFoundResponse, ProcessorPortNotFoundResponse, RegisterProcessorSourceResponse,
-    RegisteredPortResponse, RegisteredProcessorPortsResponse, RegistrationOutcome,
-    ReplaceProcessorSourceRequest, SourceProcessorPortRole, SubmittedProcessorSourceRequest,
-    UnknownProcessorTypeResponse,
+    ReplaceProcessorSourceRequest, SubmittedProcessorSourceRequest, UnknownProcessorTypeResponse,
 };
 
 /// The relative WebSocket URL carrying this runtime's live event stream — the
@@ -79,6 +72,9 @@ pub(crate) fn build_router(
     // mechanism parity, not a trust boundary the tap itself imposes. Clone the
     // token before it is moved into the mutating-route middleware below.
     let tap_auth_token = auth_token.clone();
+    // The MCP endpoint exposes the same mutating ops as tools, so it is gated
+    // exactly like the mutating routes when auth is opted in.
+    let mcp_auth_token = auth_token.clone();
 
     let mut protected = OpenApiRouter::new()
         .routes(routes!(create_processor))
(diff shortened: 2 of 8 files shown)


tato123:
## Live MCP verification — real MCP-SDK client against a running api-server

Beyond the hermetic unit tests, I booted `streamlib-runtime` (this branch, path-dep on `packages/api-server`) with the sandbox bypass, **auth OFF (default)** to also prove the access-first posture, and drove the `POST /mcp` endpoint with the **official Python MCP SDK** (`streamablehttp_client` + `ClientSession`, not hand-rolled curl — so real Streamable-HTTP/JSON-RPC protocol conformance is exercised).

```
[PASS] initialize handshake — protocol=2025-06-18 server=streamlib-api-server
[PASS] tools/list advertises all 7 veneer tools — got=['connect','graph','logs','remove_processor','replace_processor','submit_processor','tap']
[PASS] tools/call graph returns graph JSON — {"nodes":[{"id":"Pd1ac73t47aghdur7si5jen6y","type":{"org":"tatolab","package":"api-server","type":"ApiSer...
[PASS] tools/call logs returns a bounded sample (no hang) — {"requested":5,"received":0,"window_ms":500,"events":[]}
[PASS] invalid tools/call yields an in-band error, not a crash — submit_processor arguments: missing field `language`
[PASS] BONUS live submit_processor returned a well-formed MCP response — in-band error: Build orchestrator failed to materialize '@session/mcp-smoke': session deno port extraction failed

=== ALL CORE ASSERTIONS PASSED ===  (tester exit=0)
```

**What this proves live over the real wire:**
- MCP protocol handshake + version negotiation against a genuine client.
- All 7 tools discoverable with schemas.
- Read path (`graph`) returns the live graph.
- **`logs` returns within its 500ms monotonic window** (`window_ms:500`) — the tap/logs bounded-window fix works live, no idle-channel hang.
- Error path returns a clean in-band JSON-RPC error, not a transport crash.
- Write path (`submit_processor`) drives the full MCP → ops → runtime → build-orchestrator chain (runtime log shows Deno wire-vocabulary codegen staging `@session/mcp-smoke`) and returns a **clean transactional in-band error**.

The bonus submit didn't *land* only because the throwaway Deno source failed extraction — that's the already-merged #1424 submit-from-source path, not the veneer; the veneer correctly forwarded and returned the transactional error. Tester: `uv run --with mcp python mcp_smoke.py http://127.0.0.1:PORT/mcp`.

PR #1500, opened 2026-07-21
feat(examples): dynamic-reconfigure — live camera→display rewiring harness (#340)

### LINKED ISSUE #340: No example / harness for dynamic processor add/remove E2E validation
Discovered during #294 — step 5 of the plan ("Dynamic processor
add/remove: start camera-only, then add encoder + display live")
could not be executed because no example in \`examples/\` exercises
live reconfiguration of a running StreamRuntime graph.

## What's needed

A minimal example (\`examples/dynamic-reconfigure/\` or a subcommand of
an existing example) that:

1. Starts a camera-only pipeline, displays first frame.
2. Adds encoder + decoder + display processors live.
3. Removes encoder + decoder, leaves camera → display.
4. Exits cleanly.

Used as:
- Manual smoke-test for runtime reconfigure correctness.
- Target for CI once \`#293\` lands.
- Reproducer for #319 reconfigure-via-escalate work.

## Nice-to-have

- PNG sampling enabled throughout (existing
  \`STREAMLIB_DISPLAY_PNG_SAMPLE_*\` vars) so pre/post-reconfigure
  frames can be inspected.
- Timed progression so the example auto-exits after a fixed number of
  add/remove cycles.

## Reference

- plan/294 step 5
- Retest report notes this as P3 follow-up F.

tato123:
Closes #340

## Summary

Adds a standalone `dynamic-reconfigure` example demonstrating live camera→display rewiring while the runtime is running, and fixes an engine-layer bug it surfaced: a processor removal that cascades away a link which is also queued for wiring in the same commit used to hard-error with a spurious `LinkNotFound` during the WIRE phase.

- `examples/dynamic-reconfigure/`: new standalone example — starts a camera→display pipeline, then live-splices a passthrough processor in and back out while the runtime keeps running, proving disconnect/reconnect works against a started runtime.
- `runtime/streamlib-engine/src/core/compiler/compilation_plan.rs`: new `CompilationPlan::drop_link_adds_into_removed_processors` prunes queued link-adds whose endpoint processor is being removed in the same batch, before the WIRE phase runs.
- `runtime/streamlib-engine/src/core/compiler/compiler.rs`: call site invokes the new pruning step under a scoped graph read-lock, before the plan's `is_empty` early-return and before any phase executes.
- Engine test coverage: 2 new unit tests on `CompilationPlan`, plus a live-runtime test (`4cb72386`) exercising add/remove of a processor against a started runtime, and a headless splice test noted as covering add/remove only (`4a0692d9`).
- README correction (`127fc0e4`): the held-frame behavior during the mid-splice window is now accurately documented, with a trimmed doc and a renamed param.

## Test evidence

```
cargo test -p streamlib-engine --lib compiler:: → 101 passed, 0 failed (includes the 2 new unit tests)
```

Mental-revert check on the regression test: `drops_link_add_when_its_endpoint_is_removed_in_the_same_batch` asserts `links_to_add.is_empty()` after the call — a no-op method body leaves it non-empty, so the test is non-vacuous.

## E2E report

The example's mid-splice visual proof is a frame **freeze**, not continuous flow-through — see the "Review items" note below for the detail on why, and README `examples/dynamic-reconfigure/README.md` for the documented limitation.

- **info** — `runtime/streamlib-engine/src/core/compiler/compilation_plan.rs:36`: The engine fix (`drop_link_adds_into_removed_processors` + `compiler.rs` call site) correctly prevents a spurious `LinkNotFound` when a processor removal cascades away a link that is also queued for wiring in the same commit. Evidence: Read WIRE phase (`compiler.rs:441-448`): it hard-errors `Error::LinkNotFound` when `graph.traversal().e(link_id).first()` is `None`. Removals run first (`compiler.rs:192`) and petgraph `remove_node` cascades incident edges, so a link_add pointing at a removed processor would vanish before WIRE. The new method prunes exactly those. Fix is at the engine layer (not a consumer bandaid) and extends the existing `CompilationPlan` rather than adding a parallel abstraction — doctrine-compliant. Ran `cargo test -p streamlib-engine --lib compiler::` = 101 passed, 0 failed, including the 2 new unit tests. Mental-revert verified: `drops_link_add_when_its_endpoint_is_removed_in_the_same_batch` asserts `links_to_add.is_empty()` after the call, so a no-op body leaves it non-empty and fails — non-vacuous. Suggested next step: none — fix is correct and covered.
- **should-fix** — `examples/dynamic-reconfigure/README.md:32`: The example's mid-splice visual proof is a frame FREEZE, not continuous flow-through, which is a materially weaker reconfigure smoke-test than the ticket's stated scenario (start camera-only, then add encoder+decoder+display live so video keeps flowing). Evidence: `packages/debug-utilities/src/simple_passthrough.rs:12` confirms `execution = manual` — a one-shot that does not pump frames. So while spliced in, camera->passthrough->display delivers no live frames and the display holds its last frame (README lines 30-36 honestly document this). The live->held->live transition proves the disconnect/reconnect landed, but NOT that the newly-wired path carries live data — the ticket's encoder+decoder+display scenario implies a path that keeps delivering. The headless test (`dynamic_reconfigure_live_splice.rs`) and the engine fix carry the correctness weight; this is an example-quality gap. Suggested next step: consider splicing in a reactive forwarding processor (or add one to debug-utilities) so the mid-splice state shows live->live flow-through, giving positive proof the rewired path delivers frames. If deferred, record it as a known limitation on the PR body (already noted in README).
- **low** — `examples/dynamic-reconfigure/Cargo.toml:25`: The example pins `streamlib = { version = "0.7.23" }` while the workspace and every sibling standalone example pin 0.7.39. Evidence: on the branch worktree, `Cargo.toml` workspace version = 0.7.39; every other example (hello-streamlib, moq-roundtrip, jpeg-psnr, ...) pins 0.7.39; only dynamic-reconfigure pins 0.7.23. Not build-breaking (caret `^0.7.23` = `>=0.7.23,<0.8.0` is satisfied by the 0.7.39 link-patched crate), but it is a stale value inconsistent with the uniform convention. Suggested next step: bump the pin to 0.7.39 to match the workspace and sibling examples.
- **low** — `runtime/streamlib-engine/src/core/compiler/compilation_plan.rs:52`: Two near-identical `.iter().any(|removed| removed == &...)` closures where `Vec::contains` reads better and drops the redundant closure. Evidence: `let source_removed = processors_to_remove.iter().any(|removed| removed == &link.from_port().processor_id);` and the identical `target_removed` variant for `to_port()`. `ProcessorUniqueId` derives `PartialEq` (`processor_unique_id.rs:10`), so `processors_to_remove.contains(&link.from_port().processor_id)` compiles and is the idiomatic form. Suggested next step: replace both `.iter().any(|removed| removed == &X)` with `processors_to_remove.contains(&X)`; optionally collapse the pair to `[link.from_port().processor_id, link.to_port().processor_id].iter().any(|pid| processors_to_remove.contains(pid))`.
- **info** — `runtime/streamlib-engine/src/core/compiler/compilation_plan.rs:48`: `drop_link_adds_into_removed_processors` silently drops a second class of link-adds its name doesn't advertise: any queued link whose id is simply absent from the graph. Evidence: `let Some(link) = graph.traversal().e(link_id).first() else { tracing::debug!("...link absent from graph"); return false; };` — this retains-false path fires for absent-but-not-necessarily-removed links, beyond the method's stated "...into_removed_processors" scope. Defensible (an absent link can't be wired anyway, and it only debug-logs) but the behavior is wider than the name. Suggested next step: no change required; if a future reader relies on the name as the full contract, either split the absent-link filter out or widen the doc's first line to name both cases.
- **info** — `runtime/streamlib-engine/src/core/compiler/compiler.rs:170`: Call site correctly extends the existing compiler plan pipeline and scopes the graph read-lock to the minimum — no parallel abstraction, per engine doctrine. Evidence: `{ let graph = graph_arc.read(); plan.drop_link_adds_into_removed_processors(&graph); }` placed before the `is_empty` early-return and before any phase runs, with the read guard dropped at block end. Suggested next step: none — noted as positive coverage.


tato123:
## ⚠️ verify-live: FAILED — do not merge yet

Ran the example on the rig (camera `/dev/video0` vivid → display `:1`, bypass, PNG sampling, 3 cycles). It exposed a real defect; the example does **not** demonstrate live reconfigure end-to-end.

**What worked:** camera + display materialized and connected; 11 real 1920×1080 frames captured during the initial live phase.

**What failed:** at cycle 1's splice-IN the example disconnects camera→display, then adds `SimplePassthrough` from `@tatolab/debug-utilities`. The runtime spent ~21s trying to materialize that package and the **cargo build failed**:

```
21:51:38  ↳ cycle 1/3: splicing passthrough IN         (camera→display link disconnected)
21:51:59  ERROR add_v_op — '@tatolab/debug-utilities/SimplePassthrough (version-free)' is not registered — node in Error state
21:51:59  ✗ reconfigure step failed: Build orchestrator failed to materialize '@tatolab/debug-utilities':
          cargo build -p streamlib-debug-utilities exited non-zero
21:52:00  ⏹️  Reconfigure cycles complete — stopped.    (exit 0)
```

Because the direct link was already torn down and the passthrough never inserted, **frames stopped permanently** — all captured frames are from the first ~2.5s. Engine teardown confirms it: `Compile complete: +0 -3 processors` (camera+display+passthrough still present), i.e. the splice-OUT and cycles 2–3 never ran.

**Root-cause direction:** camera and display carry the *identical* `streamlib-macros = 0.7.25` dep and materialized fine — but they were built at **startup**. `debug-utilities` failed when materialized **live (post-`start()`)**. So the suspect is the **lazy/live-reconfigure materialize path**, not the package. (A startup reference to the same package should be tried to confirm.)

**Secondary bug (this example):** the run printed `Reconfigure cycles complete — stopped` and exited 0 even though a reconfigure step errored — the `Err` arm prints `✗ …` but the outer "complete" line then prints unconditionally. A failed reconfigure must be a non-success exit.

**Verdict:** #340 is not ready. Two fixes needed before merge: (1) root-cause + fix the live-materialize failure (likely engine build-orchestrator, or the example must pre-warm the package), (2) the example must not report success on a failed step.

tato123:
## ✅ verify-live PASSED — live dynamic reconfigure works end-to-end

Rebased onto current main (all of M38's compile-on-install) and re-ran on the rig (camera `/dev/video0` vivid → display `:1`, bypass, PNG sampling, 3 cycles). **The M38 compile-on-install is what unblocked this** — debug-utilities now compiles.

### First run (cold) — proved the splice + compile-on-install
```
✓ Pipeline connected → ✓ Running
↳ cycle 1/3: splicing passthrough IN
  materialized package @tatolab/debug-utilities  rebuilt=true   (compiled live, ~22s)
↳ cycle 1/3: splicing passthrough OUT
↳ cycle 2/3: splicing passthrough IN   (reused — no rebuild)
↳ cycle 2/3: splicing passthrough OUT
↳ cycle 3/3: splicing passthrough IN / OUT
26 real frames captured (vivid color bars flowing camera→display)
```
Exit was 1 on a teardown `LinkNotFound` — root-caused to an **engine** gap (not the example): petgraph's `remove_node` cascades incident edges, so a link queued for wiring against a processor removed in the *same commit* vanishes before the WIRE phase → spurious `LinkNotFound`, surfacing only at final-cycle shutdown.

### Engine fix (this branch)
`fix(engine): make processor removal idempotent to its cascaded links` — `CompilationPlan::drop_link_adds_into_removed_processors`, invoked in `compile()` before any phase, drops link-adds whose endpoint processor is removed in the same batch. Removing a processor whose links an operator already disconnected (a natural agent sequence) no longer errors. 2 revert-proof headless unit tests; 38 gates pass; `install-packages` compiled 15/15.

### Second run (with fix) — CLEAN
```
↳ cycle 1/3: splicing passthrough IN / OUT
↳ cycle 2/3: splicing passthrough IN / OUT
↳ cycle 3/3: splicing passthrough IN / OUT
⏹️  Reconfigure cycles complete — stopped.
EXIT=0            (no LinkNotFound, no error)
```
This run took ~13s (vs ~90s cold) because debug-utilities was **reused from the prior build** — compile-once-then-reuse confirmed. Frame content verified real (SMPTE color bars + vivid overlay, `1920x1080`).

**Result:** live add/connect/disconnect/remove on a running graph works across 3 cycles with a clean exit. #340's core is proven, plus an engine robustness fix that benefits every consumer.

PR #1511, opened 2026-07-22
fix(packages): debug-utilities cdylib-safe frame upload; delete build debris; stale tests

tato123:
Closes #1504

## Summary

`packages/debug-utilities` called the host-only `TextureRing::copy_pixel_buffer_to_slot`, which
does not exist on the plugin-SDK twin — the crate has been cdylib-unsafe since the engine-free
carve-out, and nothing caught it because `packages/*` are not workspace members. This PR:

- Reworks `bgra_file_source.rs` to upload frames via the sanctioned pooled-pixel-buffer producer
  pattern (`acquire_pixel_buffer` + host-visible memcpy + `pool_id.to_string()` as `surface_id`),
  mirroring the canonical h264/h265 processors instead of the host-only `TextureRing` ring.
- Adds `plugin_load_smoke.rs`, an in-crate `#[cfg(test)]` module that force-compiles the crate
  standalone against the real descriptor macro output, closing the gap that let the break go
  latent (no other CI path compiles `packages/*` in isolation).
- Deletes stale tests that imported the removed `streamlib::sdk` facade and could never compile
  under the engine-free package model (`packages/audio/tests/sdk_pubsub_e2e.rs`,
  `packages/audio/tests/sdk_pubsub_resolves.rs`, `packages/jpeg/tests/jpeg_smoke.rs` + its
  fixtures, `packages/webrtc/tests/whip_client_test.rs`), per the ticket's "fix or delete"
  authorization.
- Removes the `packages/{mavlink,network,vadr-vision}` build debris locally (untracked/gitignored
  — see review note below on why this doesn't show in the diff).

## Test evidence

- Built the branch's `debug-utilities` lib test target against the SDK twin (via the
  camera-display link config): compiles clean, `test result: ok. 2 passed` — including
  `plugin_load_smoke::simple_passthrough_descriptor_loads` and
  `bgra_file_source_descriptor_loads_cdylib_safe`.
- Negative-tested the regression guard: reverted to main's unfixed `bgra_file_source.rs` and
  recompiled against the same SDK twin — fails with `error[E0599]: no method named
  copy_pixel_buffer_to_slot found for struct TextureRing`, confirming the new smoke test is a
  real (non-vacuous) compile-time gate, not a no-op.

- **[should-fix]** `packages/debug-utilities/src/plugin_load_smoke.rs:15` — Module doc claims the
  smoke "proves every processor's macro-generated definition is well-formed," but only 2 of the
  crate's 4 Linux processors are actually introspected — `VideoFrameCounter` and
  `JpegBytesSource` are omitted (tests cover only `SimplePassthrough` and `BgraFileSource`). Either
  add descriptor introspection for the other two processors, or narrow the doc sentence to what's
  actually asserted.
- **[low]** `packages/debug-utilities/src/bgra_file_source.rs:162` — The pooled-buffer copy
  diverges from the canonical h264/h265 pattern it mirrors: no `dst_ptr` null check and no clamp
  to `plane_size(0)`. h264's decoder guards both; here `frame_size` (width*height*4) is copied
  unconditionally. Sizes are equal by construction and this matches the pre-existing unchecked
  copy, so behavior is unchanged, but it's a minor robustness gap versus the pattern being
  canonicalized. Optional: mirror h264's null-check + clamp for full parity; safe to ship as-is
  given this is a fixture-only debug source.
- **[low]** `packages/debug-utilities/src/bgra_file_source.rs:148` — Comment says holding
  `pixel_buffer` live "through `outputs.write` below so the pool can't rotate this slot out
  mid-flight," but the real in-flight safety comes from the 4-deep pool
  (`POOL_PRE_ALLOCATE_COUNT=4`) plus the `strong_count>2` skip, not from the handle's scope-end
  drop — nothing references the handle after the write, so the liveness is implicit rather than
  enforced. Not a correctness defect; optionally anchor with an explicit `drop(pixel_buffer)` (or
  reword the comment) so the ordering is visible to future edits.
- **[info]** `packages/debug-utilities/src/lib.rs` — Acceptance criterion 2 (debris dirs
  `packages/{mavlink,network,vadr-vision}` removed) isn't reflected in the diff: `git ls-tree`
  shows zero tracked files under those dirs already, and on the working tree they exist only as
  untracked/gitignored `Cargo.lock`+`target` debris. A branch commit can't remove untracked
  working-tree dirs, so this is a local cleanup operation invisible to the PR — nothing to review
  in-diff, but noting it so the acceptance box isn't mistaken for unaddressed.
- **[info]** `packages/jpeg/tests/jpeg_smoke.rs` and the other three deleted test files — all
  imported the removed `streamlib::sdk` facade and had no `[dev-dependencies]` path to compile
  under the engine-free package model (jpeg_smoke also used wall-clock
  `SystemTime`/`UNIX_EPOCH`). Deletion is within the ticket's "fix or delete" scope; the new
  descriptor smoke test uses the real `streamlib-plugin-sdk` dependency instead. Only note: JPEG
  decode behavior now has no in-package test — file a follow-up if JPEG runtime coverage is
  wanted.
- **[info]** Core fix and regression guard both verified correct — see Test evidence above for the
  full trace (`acquire_pixel_buffer` → `surface_store` registration →
  `resolve_texture_by_surface_id` → `refresh_pixel_buffer_texture`, and the non-vacuous compile
  failure on revert).
- **[low]** `packages/debug-utilities/src/plugin_load_smoke.rs:31` — the port-name extraction
  chain (`descriptor.<ports>.iter().map(|p| p.name.as_str()).collect::<Vec<_>>()`) repeats three
  times across the two test fns. A tiny `port_names()` test helper would DRY it; optional given
  it's test code with only three call sites.
- **[low]** `packages/debug-utilities/src/plugin_load_smoke.rs:4` — the module doc is a
  three-paragraph essay restating context the test names/asserts already convey; the one novel
  invariant worth keeping is "an engine-only call fails the whole crate's compile, so no processor
  can load" — the rest could be trimmed per the repo's comments rule against multi-paragraph
  rustdoc essays.


tato123:
**Verdict: APPROVE-WITH-NITS** — independent second-opinion pass; no must-fix items found.

**Correctness (verified against engine source, not just the diff):**
- The pooled-buffer rework is a faithful mirror of the canonical h264/h265 decoder pattern. `acquire_pixel_buffer` registers the `pool_id` with the surface store at acquire time (`gpu_context.rs`), so downstream `resolve_texture_by_surface_id` resolves the emitted `surface_id` — the frame path is sound end-to-end.
- No use-after-free or rotation hazard: pool buffers are permanently Arc-held by the ring + cache, and the pool's `strong_count` check skips any slot still in flight. Worst case under a badly lagging consumer is stale-content skew, which is identical to the canonical h264/h265 behavior — an engine-layer property, not something introduced here.
- The deleted tests were genuinely dead: all four imported `streamlib::sdk::*`, and no `streamlib` facade crate exists — packages depend only on `streamlib-plugin-sdk`. Nothing unrelated was deleted (the jpeg fixtures went with the only test that used them).
- Smoke test is non-vacuous: the whole crate compiling against the SDK twin is the guard, and the PR's negative test (revert → `E0599`) demonstrates it.

**Agree with the prior verifier's items:**
- Should-fix before merge: `plugin_load_smoke.rs` module doc says it proves "every processor's" descriptor, but only 2 of the 4 Linux processors are introspected (`VideoFrameCounter`, `JpegBytesSource` omitted). Either add the two descriptors or narrow the sentence.
- The four low nits (missing null-check/plane-size clamp vs the h264 copy it mirrors; the comment over-crediting handle lifetime for in-flight safety; test-helper DRY; module-doc length) — all fair, none blocking.

**One additional nit the first pass didn't state explicitly:**
- GitHub CI never compiles `packages/*` (workspace tests only; no workflow runs package tests), so the new smoke test fires only in the local/loop gate battery. The guard is real but shouldn't be mistaken for CI coverage — CI-side package compilation remains the #1502-adjacent gap.

Rules check: BUSL header present on the new file, `tracing` only, no engine-only calls remain, naming fine, scope clean.

---
*Automated second-opinion review (independent of the pre-PR verifier).*

PR #1515, opened 2026-07-22
feat(packages): package-cache build-once-reuse + disk-discipline (precursor to #1506)

tato123:
## What

The **build-cache disk-discipline + Rust build-once-reuse** foundation for #1506 — the language-agnostic infrastructure the `streamlib_modules` co-location move reuses verbatim. Landing it as a tested precursor; **this PR does NOT close #1506** (the in-place relocation follows in a dedicated migration pass).

## Contents
- **Rust build-once-reuse** (`rust_reuse_permitted` + `slot_has_host_cdylib`): an immutable/registry-pinned Rust package with a present host-triple cdylib skips the cargo re-invoke on second load; an active `link` still rebuilds. (Fixes the "Rust re-invokes cargo on every load" gap.)
- **Artifact-only retention** (`prune_build_scratch`): drops `target/` after a successful build, keeps `lib/<triple>/*.so` + `.venv/` + `_generated_/` + sidecar. Gated on `active_link.is_none()` so the dev-link loop keeps cargo incrementality.
- **`streamlib pkg cache-gc`** verb (distinct from `pkg clean`) + gitignore for co-located build output.
- **Atomic-swap hardening**: displace-to-`.old` then rename (closes the absent-slot window), with a real failure-path regression test (`atomic_swap_preserves_the_prior_slot_when_the_rename_in_fails`) that fails on the old remove-then-rename shape.

## Tests / gates
35 gates pass, 0 fail (full local battery incl. xtask lint suite + `-p streamlib …--lib`). New: `rust_reuse_permitted_*`, `slot_has_host_cdylib_*`, `prune_build_scratch_*`, `atomic_swap_preserves_*`, CLI `reclaim_*`/`sweep_*`.

## Follow-up (the #1506 core)
The in-place location move — build into version-free `streamlib_modules/@org/name/`, retire `.streamlib/cache/packages/`, sweep the single location seam across all four derivers — is a ~80-site, 5-crate all-or-nothing cutover landing as its own dedicated pass gated by the full engine test suite + /verify-live. This infra carries forward into it unchanged.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -77,6 +77,13 @@ examples/camera-python-display/assets/*
 .envrc
 .cargo/config.toml
 
+# Co-located package build output (compile-on-install): the compiled cdylib
+# and build-provenance sidecar live beside a package's source, inside its
+# `streamlib_modules/@org/name/` slot — regenerable artifacts, never committed.
+# (`.venv/`, `target/`, `_generated_/` are already ignored above/below.)
+.streamlib-build.json
+streamlib_modules/**/lib/
+
 plan.md
 
 # In-flight task handoff notes (working files for cross-session pickup,
--- a/runtime/streamlib-engine/src/core/runtime/install.rs
+++ b/runtime/streamlib-engine/src/core/runtime/install.rs
@@ -26,7 +26,10 @@ use streamlib_idents::{
 };
 
 use super::module_loader::host_target_triple;
-use super::{BuildEventSink, BuildOrchestrator, BuildPolicy, BuildRequest, BuildSource};
+use super::{
+    BuildEventSink, BuildOrchestrator, BuildPolicy, BuildRequest, BuildSource,
+    PackageSourceProvenance,
+};
 use crate::core::config::{InstalledPackageEntry, InstalledPackageManifest};
 
 /// Knobs for [`install`]. Defaults are the ordinary "install this app"
@@ -172,6 +175,7 @@ pub fn install(
         let request = BuildRequest {
             package: pkg_ref.clone(),
             source: BuildSource::PackageDir(pkg.root_dir.clone()),
+            source_provenance: provenance_of(&pkg.source),
             policy,
             host_triple: host_target_triple().to_string(),
         };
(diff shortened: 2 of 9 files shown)


tato123:
**Verdict: REQUEST-CHANGES** — one reachable correctness gap in the Rust reuse gate; everything else is solid (the atomic-swap failure-path test genuinely locks the fix, the prune/link gating reads the same link discovery the reuse gate uses, and cache-gc cannot delete a `lib/<triple>/` artifact).

## Must fix

1. **`rust_reuse_permitted` can serve a stale cdylib for a mutable-source (unlinked) Rust package whose build resolves code outside the package dir** — `tools/streamlib-build-orchestrator/src/lib.rs:160` + the gate at ~:245-254.
   The gate's premise is "no active `streamlib link` ⇒ immutable / registry-pinned source", but nothing enforces that. `Strategy::Path { path, build }` (module_loader `source.rs:330`) routes a user's mutable package dir through this same `IfStale` path, and a package Cargo.toml may carry `path = "…"` deps or workspace-inherited deps that resolve OUTSIDE `pkg_dir` — in-tree examples exist today (`packages/api-server/Cargo.toml:37` → `path = "../../sdk/streamlib-sdk"`). `compute_inputs_hash` only sees `pkg_dir`, so: build once → edit the SDK/engine → second load has an unchanged fingerprint, no link, cdylib present → **cargo is never re-invoked and the stale cdylib is served**. The pre-PR code defended exactly this (its deleted comment: "a Rust cdylib can link code outside the package dir … cargo's own fingerprint is the only correct oracle"); this PR removes the defense without a replacing guard.
   Fix shape (either works): thread source provenance into the reuse decision (registry/slpkg/git-rev = immutable ⇒ reuse eligible; `Strategy::Path` = mutable ⇒ always re-invoke cargo, which short-circuits cheaply when clean), or deny reuse when the package's Cargo.toml declares any `path` dependency / sits under a cargo workspace root outside `pkg_dir`. Provenance is the principled fix — the gate's comment already describes it as the intent.

## Nits

- **`atomic_swap` doc overclaims** (`streamlib-build-orchestrator/src/lib.rs:574-580`): "a concurrent reader now observes either the old slot or the new one, never nothing" is not true — between `rename(final → .old)` and `rename(temp → final)` the slot IS absent. The window is narrowed from O(recursive-delete) to two renames (and the failure path now restores — the real win), not closed. True closure needs `renameat2(RENAME_EXCHANGE)`. Fix the comment (and the `…_without_an_absent_window` test name encodes the same overclaim); the behavior itself is a strict improvement.
- **`sweep_staging_residue` misses the co-located modules shape it documents** (`streamlib-cli/src/commands/pkg.rs:294`, called at :251-253): it scans only DIRECTLY under `streamlib_modules/`, but slots there are `@org/name/`, so `stage_temp_dir`/`atomic_swap` residue (`.tmp-*` / `.old-*`) sits under `@org/` — one level deeper, never swept. Dead today (nothing stages there yet) but the PR states this infra carries into #1506 unchanged; recurse to slot parents or sweep alongside the `streamlib.yaml` walk.
- **`pkg cache-gc` can sweep a live in-flight build's `.tmp-*`** (`pkg.rs:294`): temp names embed a pid but the sweep has no liveness/age check. Consequence is bounded (the build fails at sidecar-write/rename; the prior slot survives via the new restore path), but a pid-liveness or mtime-age filter would make the manual verb safe to run anytime.
- **`prune_build_scratch` deletes a user-owned `target/` for `Strategy::Path` packages** (`streamlib-build-orchestrator/src/lib.rs:417`): for registry/slpkg-extracted slots pruning is clearly right, but `pkg_dir` for a path-strategy package is the user's own source dir — an unlinked local-path dev loop loses all cargo incrementality on every edit (and their manual `cargo test` state too). The same provenance flag from the must-fix resolves this: prune only orchestrator-owned source dirs.
- **`pub fn prune_build_scratch`** (`lib.rs:643`) has no caller outside the crate — keep it private until the #1506 pass actually needs it exported.

Checked and fine: `slot_has_host_cdylib` cannot false-positive on a wrong-triple `.so` (it reads only `lib/<host-triple>/`, and the sidecar triple check backs it); `compute_inputs_hash` excludes `target/`/`lib/`/sidecar so pruning never perturbs the reuse oracle; `atomic_swap_preserves_the_prior_slot_when_the_rename_in_fails` is genuinely mentally-revertible (remove-then-rename wipes the slot before the failing rename and has no restore — the assertion fails under the old shape); `reclaim_package_target_dirs` stops at the first `streamlib.yaml` and never descends into package source; CLI `println!` matches the file's existing convention; naming passes the zero-context test throughout.

---
*Automated second-opinion review (independent of the implementing agent).*

PR #1567, opened 2026-07-23
feat(cli): streamlib nodes discovery + registry-resolved control-verb targets

tato123:
**Stacked PR — base is `feat/1561-cli-control-client-verbs`, not `main`.** Merge order: `feat/1561-cli-control-client-verbs` merges first, then this PR (`feat/1561-nodes-registry-discovery`) rebases/retargets onto `main`.

Closes #1561.

## Summary

- Adds an on-disk node discovery registry in `packages/api-server` keyed to the control endpoint, so a running node advertises itself for `--node <runtime_id>` resolution.
- Adds `streamlib nodes` discovery to the CLI, and makes control-verb commands (`logs`, etc.) resolve their target via the registry when invoked with `--node` instead of an explicit `--url`.

## Test evidence

- Registry unit tests, including `scan_skips_an_entry_with_an_unrecognized_schema_version` (schema-version-mismatch handling for `scan_entries`).
- CLI-side tests for `--node` resolution and `nodes` listing, using isolated `XDG_RUNTIME_DIR` fixtures.

These were raised during independent verification of the branch. None are blocking — the change is functionally sound and ready for the owner's review — but they should be addressed before or shortly after merge.

1. **should-fix** — `packages/api-server/src/node_registry.rs:152` — `read_entry`'s doc says a version-mismatched entry is an error, but the body applies no `schema_version` guard — it decodes and returns `Ok(Some(entry))` for any `schema_version`. Confirmed self-review finding (a): a latent inconsistent-reader bug that bites when schema v2 lands, since `read_entry` is exactly the path `--node <runtime_id>` resolution uses (`control.rs` `resolve_control_url`).
   - Evidence: `read_entry`: `let entry: NodeRegistryEntry = serde_json::from_slice(&bytes).map_err(...EntryDecode...)?; Ok(Some(entry))` — no `entry.schema_version != NODE_REGISTRY_SCHEMA_VERSION` check, whereas `scan_entries` (same file) has one and is tested by `scan_skips_an_entry_with_an_unrecognized_schema_version`. Doc comment: "A present-but-corrupt or version-mismatched entry is an error." No test exercises `read_entry` against a version mismatch, so the gap is unguarded.
   - Suggested next step: Add the same `schema_version` guard to `read_entry` (return an `EntryDecode`/version error), and add a `read_entry` version-mismatch test that fails against the current unguarded code; or correct the doc to state `read_entry` does not version-check. Prefer the guard to keep `--node` resolution strict.

2. **should-fix** — `tools/streamlib-cli/src/main.rs:46` — `logs --count` dropped its `requires = "url"` (removed to allow control mode via `--node`), so in LOCAL mode `streamlib logs <runtime_id> --count 5` is now silently accepted and ignored — `LogsArgs` has no `count` field. Confirmed self-review finding (b): a UX regression from the base, which rejected the flag at the clap layer.
   - Evidence: Diff: `#[arg(long, value_name = "COUNT", requires = "url")]` → `#[arg(long, value_name = "COUNT")]`. Local path in `async_main` dispatches to `commands::logs::run(LogsArgs{..})` which has no `count` field (`logs.rs:17-28`); `count` is threaded only through the control-plane branch `commands::control::logs(&url, count)`. clap `requires` can't express url-OR-node, so the guard was dropped wholesale rather than re-expressed.
   - Suggested next step: Re-express the constraint as an `ArgGroup` (count requires one of url|node) or error at runtime when count is present in local mode, so the flag is never silently ignored.

3. **should-fix** — `tools/streamlib-cli/src/commands/nodes.rs:141` — `control_url_reachable()` hand-rolls the MCP request seam (`{url}/mcp` endpoint construction, `STREAMLIB_MCP_TOKEN` bearer-header logic, and the `tools/call`/`graph` JSON-RPC envelope) that `control::post_mcp_request` already owns and whose own doc declares it "the single request/response seam ... every control verb share". This new sibling module in the same crate bypasses that seam, duplicating the endpoint+token+envelope construction — engine-doctrine "search first, extend never parallel".
   - Evidence: `nodes.rs:146` `let endpoint = format!("{}/mcp", control_url.trim_end_matches('/'));` and `:157` `if let Ok(token) = std::env::var("STREAMLIB_MCP_TOKEN")` + the jsonrpc envelope at `:148` duplicate `control.rs` `post_mcp_request` (`:92` identical `/mcp` construction, `:95` identical bearer header) and `call_tool`'s envelope (`:279`). The only real divergence is the short timeout and treating an HTTP status error as reachable.
   - Suggested next step: Give `post_mcp_request` an optional timeout param (or add a `probe_reachable(url, token) -> bool` beside it that treats `ureq::Error::Status` as alive) and have `control_url_reachable` call through it, so `/mcp`+token+envelope live in one place.

4. **should-fix** — `packages/api-server/src/node_registry.rs:155` — `read_entry`'s doc comment states "A present-but-corrupt or version-mismatched entry is an error", but the body never compares `entry.schema_version` against `NODE_REGISTRY_SCHEMA_VERSION` — a future-version entry deserializes fine and is returned `Ok(Some(..))`. This both breaks the stated contract and diverges from `scan_entries` (`:200`), which does skip version-mismatched entries. The `--node <runtime_id>` resolver (`control.rs` `resolve_control_url`) relies on `read_entry`, so a mismatched wire version silently resolves and gets used while `nodes` listing hides it — inconsistent handling of the same wire-version concern.
   - Evidence: `node_registry.rs:162-164` decodes then `Ok(Some(entry))` with no `if entry.schema_version != NODE_REGISTRY_SCHEMA_VERSION` guard; contrast `scan_entries` at `:200-206` which warns and continues on mismatch. No test covers `read_entry` against a mismatched version.
   - Suggested next step: After decode in `read_entry`, reject on `entry.schema_version != NODE_REGISTRY_SCHEMA_VERSION` (add/reuse an error variant), and add a regression test — or correct the docstring if pass-through is intended.

5. **low** — `tools/streamlib-cli/src/main.rs:47` — Dropping the previous `requires = "url"` from `Logs --count` (to also permit `--node`) leaves it with no gate at all, so `streamlib logs <runtime_id> --count 5` in local-file mode now silently accepts and ignores `--count` (local `LogsArgs` has no `count` field). Previously clap rejected it.
   - Evidence: `main.rs:47` `#[arg(long, value_name = "COUNT")] count: Option<usize>` — no requires/conflicts; local `commands::logs::run` never receives `count` (`logs.rs` `LogsArgs` has no `count` field).
   - Suggested next step: Model "requires url OR node" with a clap `ArgGroup` (`requires = "target"`) rather than dropping the requirement entirely, so `--count` without a control target is rejected.

6. **low** — `tools/streamlib-cli/src/commands/control.rs:68` — `render_node_hint`'s doc says it emits a " (candidates: ...)" fragment, but the format string emits " Live nodes: {listed}.". Doc drifts from the string it documents.
   - Evidence: `control.rs:68` doc "A trailing `(candidates: ...)` fragment" vs `:80` `format!(" Live nodes: {listed}.")`.
   - Suggested next step: Align the doc wording with the actual "Live nodes:" fragment.

7. **info** — `tools/streamlib-cli/src/commands/control.rs:625` — `with_isolated_node_registry` (`control.rs` tests) and `with_isolated_xdg_runtime_dir` (`node_registry.rs` tests) are near-identical unsafe `XDG_RUNTIME_DIR` env-swap-and-restore helpers. Test code is exempt and they sit in different crates (`streamlib-cli` vs `api-server`), so extraction isn't obviously worth a shared test-support surface — noting only as an observation.
   - Evidence: Both wrap `set_var`/`remove_var` under an `unsafe` block guarded by `#[serial]`, restoring the prior value; only the closure signature differs.
   - Suggested next step: No action unless a third copy appears; if it does, consider a `#[cfg(test)]` helper in a shared test-support module.


tato123:
## Live rig verification — discovery → drive → teardown ✅

Ran a real end-to-end smoke on the rig (built binaries, sandbox bypass for the IPC/control-plane runtime) against `streamlib-runtime` — the actual control-endpoint node (it statically hosts the `ApiServer` processor). This exercises this PR's registry+discovery+resolver **and** #1566's `graph` verb over the live `/mcp` seam.

**1. `streamlib nodes` before start** — empty, correct invariant note:
```
No running nodes found in /run/user/1000/streamlib/nodes.
(Only runtimes hosting an ApiServer control plane appear here.)
```

**2. Registry entry written when the ApiServer processor bound its port** (`$XDG_RUNTIME_DIR/streamlib/nodes/<runtime_id>.json`, actual bound port):
```json
{
  "schema_version": 1,
  "runtime_id": "Rbelp7txygov2r62opqsnpyrr",
  "control_url": "http://127.0.0.1:9210",
  "pid": 2534163,
  "hint": "streamlib-runtime (.../.claude/worktrees/wf_d0287dd1-186-2)"
}
```

**3. `streamlib nodes`** — discovered + liveness-checked:
```
RUNTIME_ID                 CONTROL_URL                PID  ALIVE?  HINT
Rbelp7txygov2r62opqsnpyrr  http://127.0.0.1:9210  2534163  yes     streamlib-runtime (...)
```

**4. `streamlib graph`** (no `--url`/`--node` → sole-live-node resolver) — returned the live graph over `/mcp`, pretty-printed (ApiServer processor, state `Running`, config shows bound port 9210). `streamlib graph --node Rbelp7txygov2r62opqsnpyrr` returned the identical graph, confirming registry resolution.

**5. Teardown** — SIGTERM → runtime exited → entry removed on clean teardown; `streamlib nodes` empty again, registry dir clean.

### Parked (not covered by this live run)
The media-pipeline verbs (`tap` of real `camera/frames`, `submit`/`replace` hot-swap) are covered by unit tests but not this live run: the issue's acceptance walkthrough starts from `cargo run -p camera-display`, but **camera-display hosts no `ApiServer` processor** (its graph is Camera → Display only), so it writes no registry entry and isn't discoverable — by this PR's own (correct) design that ties entries to a control endpoint. Raised as an owner decision on #1561.

tato123:
## Full live walkthrough ✅ (supersedes the "parked media steps" note above)

My earlier "camera-display isn't drivable" framing was wrong — the api-server is statically linked into `streamlib-runtime`, and modules resolve via `streamlib link` off the filesystem (no registry). Re-ran the real acceptance walkthrough end-to-end on the rig against `streamlib-runtime --snapshot <camera→display>`, anchored in the engine-linked `examples/camera-display` app dir:

- **Module resolution via `streamlib link`** (no registry): the runtime resolved `@tatolab/camera`, `@tatolab/display`, and their `@tatolab/core` schema dep from the linked checkouts, building `@tatolab/display` on demand — `streamlib link active — resolving module from linked checkout (overriding strategy)`.
- **`streamlib nodes`** — the runtime listed, alive, hint = `examples/camera-display`.
- **`streamlib graph --node <id>`** — full live topology `[ApiServer, Camera, Display]` with the `camera.video → display.video` link, over `/mcp`.
- **Real frames flowing** — camera log: `opened 'vivid'`, `capturing 1920x1080 NV12`, `first frame captured via GPU compute`, … `Stopped (43 frames)`. The camera→display pipeline ran live on the GPU.
- **Teardown** — SIGTERM removed the registry entry; `streamlib nodes` empty again.

### Finding: `tap` on a GPU-video channel
`streamlib tap <camera_id>/video` returns the engine's error `no tappable channel … a channel's iceoryx2 data service exists only once a connect() has wired its source output port`. The camera→display video flows as **GPU DMA-BUF surfaces**, not iceoryx2 byte bags, so there's no byte-channel data service for `tap` to attach to. The `tap` verb is a faithful thin client (it connects, forwards, and surfaces the engine's exact error); this is pre-existing engine tap behavior, not a #1561 CLI defect. Worth a note that the issue's "tap camera/frames → pixels" step needs a byte-bag channel (or a `tap` enhancement for GPU-surface video) — flagged separately.

(Related cleanup filed as #1570 — the stale "registry" strings the runtime still logs even when `link` overrides.)

tato123:
### Correction on the `tap` note above

Disregard my framing of the GPU-video tap as a "finding/gap." `tap`'s purpose is to show the **bags** at a channel — the byte data flowing through it — not to render frames; rendered pixels would come from a processor that converts bags → images. So the camera→display path simply having no byte-bag channel to tap is expected, and `tap` reporting "no tappable channel" there is correct behavior, not a defect. Bag-level tapping is the flexible, reactive-system-shaped design and is working as intended.

Net: nothing to change on `tap`. A clean live `tap` demo just needs a channel that actually carries bags (an inter-processor byte-shaped channel), rather than the GPU-surface video path.

PR #1629, opened 2026-07-28
feat(engine): shutdown-request primitive so a harness can own the run loop

tato123:
Closes #1599.

## What

A harness that owns the run loop can now be asked to shut the runtime down, from every surface that can reach the runtime: in-process Rust, a cdylib plugin across the plugin ABI, HTTP, MCP, and the CLI.

One funnel, one latch, two arms — `runtime/streamlib-engine/src/core/runtime/runtime_shutdown_request.rs`:

- **Host arm** (engine statically linked): set the process-local latch, then publish `RuntimeEvent::RuntimeShutdown`.
- **cdylib arm** (engine-free plugin image): publish ONLY, over the plugin ABI. A plugin image must never latch a flag the host's run loop reads — the latch lives in the host's copy of the engine, so a cdylib-local latch would be a second, invisible flag. `a_plugin_image_never_observes_its_own_shutdown_request` locks that.

The observe side is one shared predicate, `runtime_shutdown_observed`, used by both `#[cfg]` arms of `Runner::wait_for_signal_with`. `Runner::start` takes the latch at entry so a second in-process run cannot inherit the first run's request. `Runner::stop()` stays off the `RuntimeOperations` trait — stopping is the harness's call; requesting is everyone's.

The old `request_runtime_shutdown_from_plugin_abi_boundary` is deleted outright (repo-wide grep returns nothing).

## Plugin ABI: no ABI change

`git diff $(git merge-base HEAD origin/main) HEAD -- runtime/streamlib-plugin-abi/src/` is **zero lines**.

- `STREAMLIB_ABI_VERSION` stays `6` (`runtime/streamlib-plugin-abi/src/lib.rs:106`, pinned at `:1056`).
- `RUNTIME_OPS_VTABLE_LAYOUT_VERSION` stays `3` (pinned at `:1066`); its fold into `PLUGIN_ABI_LAYOUT_FINGERPRINT` (`:253`, `:263`) is untouched. No already-built plugin package is invalidated.
- No new `#[repr(C)]` type crosses the plugin ABI, so there is no layout-regression test to add and no `*_VTABLE_LAYOUT_VERSION` to bump.

This is **not** a vtable-slot change. The request rides the already-reserved control topic `PUBSUB_CONTROL_TOPIC_RUNTIME_SHUTDOWN_REQUEST = "control:runtime-shutdown-request"` (`lib.rs:442`, value locked at `:1250-1258`). The host matches it inside the `run_host_extern_c` panic net and **before** the general `Event` decode (`runtime/streamlib-engine/src/core/plugin/host_services/mod.rs:844-874`), with a warn-and-drop for any other `control:`-prefixed topic so the reserved namespace is host-defended rather than prose. `docs/architecture/plugin-abi.md` gains that namespace contract.

The reason is decoded with `unwrap_or_default()` — a bad/empty payload still shuts the runtime down, only attribution is lost. That leniency is locked by three tests (`host_pubsub_publish_still_requests_shutdown_on_an_empty_payload`, `..._on_an_undecodable_reason`, `host_pubsub_publish_drops_unhandled_reserved_control_topic`).

## Surfaces

- `RuntimeOperations::request_runtime_shutdown` — five-impl sweep (`operations.rs`, `operations_runtime.rs`, `runtime_ops_shim.rs`, two api-server stub impls).
- SDK parity: `sdk/streamlib-sdk/src/lib.rs:166-173` mirrors `sdk/streamlib-plugin-sdk/src/lib.rs:192-194`.
- `POST /api/runtime/shutdown` on the protected route group; MCP `shutdown` tool (tool count 7 -> 8, with the four-site count sweep).
- `streamlib shutdown [--url|--node] [--reason <text>]`.
- The macOS `NSApplication` loop now reads the shutdown flag. It never did — which made the already-shipped engine-free leg a **silent no-op on macOS**. That loop has no shutdown-flag hook of its own, so the observation rides in on the periodic callback: `ControlFlow::Break` -> `app.terminate(None)` -> `applicationWillTerminate` -> the existing stop callback. `apple/runtime_ext.rs` itself is untouched.

## Three things worth calling out explicitly

1. **`impl From<&HostServices> for HostCallbacks`** (`host_services/mod.rs:330-368`) plus a `#[cfg(test)] host_services_test_support` module. This is a DRY extraction of the field-by-field cache that was inline in `install_host_services`. Called out per engine doctrine: it replaces real duplication, it is not a silent refactor.
2. **A twin-drift guard** pairing the engine's `publish_runtime_shutdown_request` with the plugin SDK's twin (`twin_drift_guard.rs:202-232`), with `extract_marked_region` generalized to take begin/end markers. The guard is load-bearing precisely because one-sided drift here fails *silently*: the host decodes the reason leniently and shuts down anyway, so a drifted twin loses attribution without any error.
3. **`serve_stdio_jsonrpc` grew a fourth parameter** `stop_serving_when_shutdown_observed` (`packages/api-server/src/mcp.rs:207-231`) — a breaking signature change on a public api-server fn. The CLI side polls the latch at 100 ms (`tools/streamlib-cli/src/commands/mcp.rs:66-72`). This is why the PR is titled `feat` and not `refactor`. Also newly an observe site: `core/utils/loop_control.rs:59-66`.

## Validation

All run in the worktree, all green:

- `cargo check` on the 5 touched crates `--all-targets` — clean.
- `cargo test -p streamlib-engine --lib runtime_shutdown` — 11 passed.
- `cargo test -p streamlib-engine --lib loop_control` — 4 passed.
- `cargo test -p streamlib-engine --test runtime_shutdown_request_ends_run_loop` — 2 passed. Real `Runner`, processor-driven and latched-before-subscribe, both reaching `RuntimeStatus::Stopped` inside a 5 s watchdog. Sandbox-only, no rig.
- `cargo test -p streamlib-api-server --lib` — 56 passed.
- `cargo test -p streamlib-cli --bin streamlib` — 90 passed.
- `cargo clippy` on the 5 crates — 0 errors.
- `cargo run -p xtask -- lint-logging` — 607 files scanned, no violations.

## macOS cross-compile gate

Doctrine requires `cargo check --target aarch64-apple-darwin` for any Apple-path edit. This machine has no macOS cross toolchain, so the check was made runnable with a throwaway zig-backed shim (nothing in the repo changed):

```
zig 0.14.1 unpacked to /tmp; a cc-rs shim that strips -arch/-target/-mmacosx-version-min
and appends -target aarch64-macos.11.0.0, exported as CC_/CXX_/AR_/RANLIB_aarch64_apple_darwin;
BINDGEN_EXTRA_CLANG_ARGS="--target=aarch64-apple-macosx11.0.0 -nostdinc
  -isystem <zig>/lib/include -isystem <zig>/lib/libc/include/any-macos-any"
```

With that, `ring`, `lzma-sys`, `bzip2-sys`, `zstd-sys` and `iceoryx2-pal-posix` all build for the target and the check reaches `streamlib-engine`'s own sources.

**Result: this branch introduces zero new `aarch64-apple-darwin` errors.** The error set on this branch and on `origin/main` is byte-identical except for one line-number shift of the same pre-existing error (`host_services/mod.rs` `E0433 streamlib_adapter_abi` moves 595 -> 719 because this branch inserts 124 lines above it).

The absolute check cannot pass, on this branch or on main: **`streamlib-engine` does not compile for `aarch64-apple-darwin` on `main` today** — 39 errors, all in files this branch does not touch. Reported, not fixed (no auto-fixing unrelated issues): `extern blocks must be unsafe` across `apple/corevideo_ffi.rs`, `apple/xpc_ffi.rs`, `apple/pixel_transfer.rs`, `apple/vimage_ffi.rs`, `apple/{audio_clock,media_clock,time,thread_priority,texture_pool_macos}.rs`, `metal/rhi/gl_interop.rs`, `core/display_info.rs`; plus unresolved `rspirv_reflect` / `winit` / `streamlib_adapter_abi` / `VulkanLayout`, a missing `to_mtl_pixel_format`, an orphan `impl` in `metal/rhi/pixel_format.rs`, and missing `Arc` / `Error` imports in `host_services/gpu_context/limited/command.rs` and `core/context/texture_ring.rs`.

The macOS runtime behavior itself (the `NSApplication` terminate path) is still unverified on a real device.

## Polyglot reach: control-plane only, by construction

This PR gives the shutdown request a **language-agnostic control plane** — `POST /api/runtime/shutdown` and the MCP `shutdown` tool — which is what serves a Python or Deno *entry*. It deliberately does **not** give a Python or Deno *subprocess processor* an in-graph path: the escalate op set is the only subprocess to host request channel, and adding an op there means a schema change, generated-type regeneration, two SDK surfaces, and two rig-gated tests. Per the owner's decision on #1599 ("control plane only, and file the escalate-bridge op as a follow-up"), that is filed as #1634 into the same milestone.

## Why the control topic and not a `RuntimeOpsVTable` slot

The request crosses on the reserved `control:` plugin-ABI topic rather than a new vtable slot. A slot buys a return value or a typed error; a shutdown request is fire-and-forget with no completion payload, so a slot would cost a `*_VTABLE_LAYOUT_VERSION` bump and invalidate every already-built plugin for nothing. Nothing here grows a vtable: no slot added, no layout version moved, `HOST_SERVICES_LAYOUT_VERSION` and `abi_layout_fingerprint` untouched.

## New engine to SDK twins

`twin-guard(runtime-shutdown-publish)` lands here — a marked-region entry in `twin_drift_guard.rs` covering the cdylib-side publish, so a facade package and an engine-free package hand the host the same `(reserved topic, msgpack reason)` pair. Unbypassable: there is no fixture to update.

## Reported, not fixed

- The new test surface does not gate in CI. `.github/workflows/test.yml:70-72` runs only `-p streamlib/-macros/-plugin-abi/-plugin-sdk/-processor-extract --lib` plus the engine attribute-macro test; engine lib tests are a tracked follow-up per the workflow's own comment at `:60-69`, and api-server / CLI tests are outside CI entirely.
- `cargo fmt --all --check` is a false signal in this checkout — it flags ~110 untouched files and the same diff reproduces against `origin/main` content (local rustfmt 1.8.0 vs committed import ordering). There is no fmt gate in `test.yml`.
- `cargo clippy --all-targets` has pre-existing `disallowed_macros` errors in `runtime/streamlib-engine/tests/polyglot_linux_execution_modes.rs:78,96,100,104`.
- `reinstall_signal_handlers` (`core/signals.rs`) is dead code with no caller in the workspace — pre-existing, untouched here.

## Filed follow-ups

- #1634 — `request_runtime_shutdown` escalate op, so a Python/Deno subprocess processor gets an in-graph path (milestone 39).
- #1635 — restore `aarch64-apple-darwin` compilation (39 pre-existing edition-2024 `extern` errors), make the doctrine gate reproducible, then verify the macOS `NSApplication` terminate path on a real device.
- #1636 — emit the OpenAPI spec from the real `ApiDoc` instead of the hand-maintained duplicate in `bin/generate_openapi.rs`, which is already drifted and would ship a TS/Deno client without the shutdown endpoint.

## Round 1 self-review

This PR was squash-merged before the team's round-1 self-review finished. Every finding from that round is addressed in the follow-up **#1637**, which corrects three defects in what merged here — the POSIX shutdown `reason` reading `posix signal 88` for every signal, `Runner::start` discarding a request issued before the run loop (the start-script case #1599 names), and the per-call wrong-image warn on a polled predicate — plus the plugin-ABI seam: the `HostServices` to `HostCallbacks` field map becomes a guarded engine to SDK twin, and the public `From` impl becomes a private named constructor so the load handshake's version-first ordering is structural.


tato123:
Round-1 self-review landed after this PR was squash-merged (65ebed54), so the fixes ship as a follow-up: #1637.

Three defects it corrects in what merged here:
- the POSIX shutdown `reason` said `posix signal 88` for every signal (`pipe::register` writes a fixed `b"X"` byte, not the signal number);
- `Runner::start` took the latch at entry, so a start-script aborting from `setup(rt)` got `Ok(())` and a runtime that never stopped — the exact case #1599's "a start-script can request shutdown" names;
- `is_runtime_shutdown_requested` warned on every call inside a plugin image, and `shutdown_aware_loop` polls it once per unthrottled iteration.

Plus the plugin-ABI seam: the `HostServices` → `HostCallbacks` field map is now a guarded engine↔SDK twin (it existed twice, held by nothing), and the public `From` impl is a private named constructor so the load handshake's version-first ordering is structural rather than conventional.

Follow-ups filed from this review: #1634 (escalate shutdown op for Python/Deno subprocess processors), #1635 (aarch64-apple-darwin compilation + real-device macOS verification), #1636 (OpenAPI generator drift).

Issue #39, opened 2025-10-28
[UI] Auto-generated VST parameter web UI


tato123:
## Goal

Build **auto-generated web UI for VST plugin parameters** - no manual UI design per plugin. The system introspects VST parameters and generates appropriate controls (sliders, toggles, dropdowns).

## Why Auto-Generated UI?

VST plugins expose full parameter metadata (#37, #38), so we can:
- ✅ **Zero manual work** - UI generates automatically from plugin metadata
- ✅ **Works with any plugin** - No custom code per VST
- ✅ **Real-time sync** - WebSocket updates when agents change parameters
- ✅ **Remote control** - Control helmet audio from command center
- ✅ **Consistent UX** - All plugins use same UI patterns

### 1. Parameter Introspection API

```rust
// REST API endpoint
GET /api/processors/{processor_id}/parameters

Response:
{
  "processor_id": "reverb_1",
  "processor_name": "Valhalla VintageVerb",
  "parameters": [
    {
      "id": 0,
      "name": "Room Size",
      "short_name": "Size",
      "units": "%",
      "min": 0.0,
      "max": 100.0,
      "default": 50.0,
      "current": 75.0,
      "kind": "Continuous"
    },
    {
      "id": 1,
      "name": "Damping",
      "units": "%",
      "min": 0.0,
      "max": 100.0,
      "current": 30.0,
      "kind": "Continuous"
    },
    {
      "id": 2,
      "name": "Enabled",
      "units": "",
      "current": 1.0,
      "kind": "Toggle"
    }
  ]
}
```

### 2. Auto-Generated Controls

JavaScript automatically renders appropriate control types:

```javascript
function renderParameter(param) {
  switch (param.kind) {
    case 'Toggle':
      return renderToggle(param);
    
    case 'Discrete':
      return renderDropdown(param);
    
    case 'Continuous':
      return renderSlider(param);
  }
}

function renderSlider(param) {
  const container = document.createElement('div');
  container.className = 'parameter-control';
  
  container.innerHTML = `
    <label>${param.name}</label>
    <input 
      type="range" 
      id="param-${param.id}"
      min="0" 
      max="100" 
      value="${param.current}"
      data-processor="${param.processor_id}"
      data-param-id="${param.id}"
    />
    <span class="value-display">${formatValue(param)}</span>
  `;
  
  // Wire up event handler
  const slider = container.querySelector('input');
  slider.addEventListener('input', (e) => {
[...]

PR #1640, opened 2026-07-28
feat(extract): root Rust processor discovery at processors/ and generate the crate root

tato123:
Closes #1589.

A Rust processor author writes the struct and nothing else. `processors/` is the
one discovery root for every language and every crate-type; the crate root is
generated from it, and the same scan feeds both the generated `export_plugin!`
register list and the derived `processors:` manifest — so a hand-written `mod`
list can no longer disagree with what the manifest says ships.

## Why / what / alternatives — the two load-bearing new shapes

**`streamlib_processor_extract::crate_root`** — a generation subsystem, not a
new abstraction beside an existing one. It is driven entirely by the existing
module-reachability walk in the same crate (`reachable.rs`), which already knew
how to resolve `processors/` the way `rustc` does. Alternatives considered:
(a) generate from a `build.rs` — impossible, cargo resolves `[lib] path` at
*target resolution*, before any build script runs; (b) keep a committed
`src/lib.rs` and codemod it — leaves the hand-written list as the source of
truth, which is the defect; (c) a proc-macro `include_processors!()` — a macro
cannot enumerate the filesystem deterministically across a `.slpkg` rebuild on
a consumer's host. Generation is opt-in per package, keyed on the declared
`[lib] path`, so a crate that commits its own root is never overwritten.

**`export_plugin!`'s per-entry `#[cfg]` + first-surviving-entry anchor** — one
invocation now describes every target the package builds for, because the
generator mirrors each arm's authored `#[cfg]` verbatim rather than re-deriving
a platform rule. The anchor (which entry supplies the declaration's single
`abi_layout_fingerprint` / build identity) becomes target-dependent where it
used to be syntactically fixed. That is safe only because every entry resolves
both consts against the same SDK crate — and that agreement is now *enforced*,
not assumed: a per-entry `const _: () = assert!(...)` makes a cdylib mixing
facade-built and engine-free-SDK-built processors an E0080 at the invocation
instead of a declaration whose fingerprint describes one entry. Alternatives:
(a) one `export_plugin!` per cfg arm — `#[unsafe(no_mangle)] static
STREAMLIB_PLUGIN` can only be defined once; (b) gate the whole invocation on a
re-derived platform predicate — that is the "scanner interprets a folder
convention" shape the ticket rules out. `PluginDeclaration` is byte-unchanged
(40/8, `abi_version` still at offset 0 and read first).

## Registration-surface changes on non-Linux targets

Both are consequences of mirroring each arm's own `#[cfg]` instead of the
author's hand-written outer gate, both are net-more-correct, and both are
pinned by tests in `real_package_reachability.rs`:

- **`@tatolab/camera`** previously gated its whole `export_plugin!` on
  `#[cfg(target_os = "linux")]` — forced by the old macro shape, which had no
  per-entry `#[cfg]`, so one Linux-only entry (`CameraProcessor`, behind a
  `#[cfg(target_os = "linux")] pub mod camera`) gated the whole invocation.
  `CameraToCudaCopy` is platform-free, so on macOS/iOS the cdylib now exports a
  `STREAMLIB_PLUGIN` symbol where the loader previously failed with a
  missing-symbol error; the processor now fails in `setup()` with its own
  configuration error instead. That is the manifest's own claim honored:
  `packages/camera/streamlib.yaml` declares `CameraToCudaCopy` with no platform
  constraint, and `processors/camera_to_cuda_copy.rs` compiles its documented
  Apple stub unconditionally — gating the arm back to Linux would make that stub
  dead code and leave the Apple cdylib disagreeing with its own manifest.
- **`@tatolab/audio`** previously gated on `any(linux, macos, ios)`, so a
  Windows build registered 0 processors even though 5 of the 7 are
  platform-free. Those 5 now register.

**`@tatolab/display`** is the counter-case and keeps its refusal: its Apple
`compile_error!` comes back as its own arm
(`processors/apple_unsupported.rs`, file-gated on the Apple targets, declaring
no processor), so an Apple build still fails at compile with an actionable
message rather than producing a cdylib that registers nothing.

## Stated exceptions

- **`packages/api-server`** authors its processor under `processors/` like
  every other package, but keeps a committed `src/lib.rs` crate root: it is a
  statically-linked host rlib with a `[[bin]]` and a `src/` tree, not a
  distributable cdylib. Generation is opt-in on `[lib] path`, so the generator
  never touches it. Its committed `processors:` stays backed by code (pinned by
  `the_host_rlib_package_still_derives_its_processor_from_the_shared_root`).
- **`tools/streamlib-cross-rustc-fixture`** deliberately keeps a hand-written
  `export_plugin!` — it is a cross-rustc ABI fixture, not a package the sweep
  covers.
- **`packages/test-fixtures-abi-mismatch`** keeps no `export_plugin!` at all:
  it hand-crafts the `#[unsafe(no_mangle)] pub static STREAMLIB_PLUGIN`
  directly, deliberately bypassing the macro (which would always emit an
  agreeing `abi_layout_fingerprint`) because a *mismatching* declaration is the
  fixture's whole purpose. Against the ticket's "no package retains a
  hand-written `export_plugin!`", the count across `packages/` is zero.
- **No per-package Rust façade.** With the crate root generated there is
  nowhere to author a flattening `pub use`, so a cross-package Rust path names
  the `processors/` file layout (`streamlib_audio::processor_audio_converter::…`).
  Only two in-tree consumers exist and both are migrated. If a stable façade is
  wanted later, the cheapest seam is for the generator to `include!` an optional
  package-authored re-export file.

## Verified on this box

Both workflows added the `cargo xtask generate-crate-roots` step; both were run
here against the merged tree, not deferred to CI.

- **`install-packages`** — `cargo run --locked -p xtask -- install-packages`
  (the workflow's own ground-truth enumerator) compiled 15/15 distributable
  packages through the real `streamlib link --engine` + `streamlib add` path,
  emitting cdylibs into a scratch consumer's `streamlib_modules/`. The other
  four (`api-server`, `clap`, `test-fixtures`, `test-fixtures-abi-mismatch`)
  skip as non-distributable. This is the leg that actually `rustc`s the
  generated crate roots.
- **`check-pack-load`** — `cargo test --locked -p streamlib-pack`
  (66/3/3/1/3/0 passed) and `cargo test --locked --test pack_then_load_smoke
  -p streamlib-engine` (3 passed, 1 ignored). Its Rust-impl leg is `#[ignore]`d
  pending #1323 / #1338, so this workflow compiles no generated crate root —
  `install-packages` is the leg that does.

## Not verified here

- **Apple cross-compile** (`cargo check --target aarch64-apple-darwin`) —
  tracked by #1635, which owns restoring that gate; it is unsatisfiable on this
  rig for reasons predating this branch. The migrated Apple arms live in
  out-of-workspace packages that resolve their SDK deps from the local registry,
  and the workspace's own Apple cross-check fails on native C toolchain deps
  (`ring`, `zstd-sys`, `iceoryx2-pal-posix`, and `lzma-sys`, whose `cc-rs`
  invocation rejects `-arch` / `-mmacosx-version-min` on the host toolchain).
  Needs a runner with the Apple toolchain.

  The unverified surface is `packages/audio/processors/audio_capture_apple.rs`
  and `audio_output_apple.rs` (2 arms) and
  `packages/clap/processors/` (6 arms, every file `#![cfg(any(target_os =
  "macos", target_os = "ios"))]`). No CI leg compiles `packages/clap` on any
  target either — `install-packages` skips it as non-distributable (path dep on
  `streamlib-audio`). Every Apple edit is a file move plus the arm's own
  `#![cfg]` header, with one exception: `clap_effect.rs` now names
  `streamlib_audio::processor_audio_converter::{ProcessorAudioConverter,
  ProcessorAudioConverterTargetFormat}`, forced by the loss of the hand-written
  façade. Confirmed statically — the generated audio root declares `pub mod
  processor_audio_converter;`, both types are `pub` in that file, and every
  other `crate::…` path in `packages/clap/processors/` names a module the
  generated root declares.
- **GPU / dlopen runtime** — the 21 `load_project_dylib_*` binaries route
  through the shared `build_folder_backed_package` helper and need the rig.

## Follow-up fixes on this branch

**Per-platform routing shims are gone.** The `src/` → `processors/` move carried
`processors/{apple,linux}/mod.rs` across intact — a file-level `#![cfg]` plus
`pub mod` + `pub use`, with the implementation a level down. That shim only ever
existed because the old `export_plugin!` needed one Rust path per entry; with
per-entry `#[cfg]` there is nothing left for it to unify. Platform variants are
now sibling files under `processors/`, each declaring the **same processor id**
behind its own mutually-exclusive `#![cfg]` (`audio_capture_linux.rs` /
`audio_capture_apple.rs`), and the generated root projects that directly. Seven
packages had a shim: audio (both arms), camera, display, h264, h265, jpeg, mp4.
`screen-capture` is fully parked and `webrtc/processors/streaming/` is an
ordinary nested support module with no `#![cfg]` — neither needed a change.
Nested directory arms stay supported; the two generator fixtures that exercised
them keep their coverage under a non-platform name (`processors/capture_backends/`)
rather than modelling the shim they document away. Every deleted `pub use`
re-export had zero consumers.

**The empty-scan refusal was load-bearing at only one of its three call sites.**
`check-cdylib-reach` was correct; `lint-logging` summed one `files_scanned`
across the Python SDK, the Deno SDK and the Rust AST walk, and
`check-processor-spec-new` counted every `.rs` its walk read while its
hand-rolled-`SchemaIdent` pass only fires on example sources a source root
selects. Either way one set going to zero hid behind the others. Reproduced
empirically rather than by reading — the code reads correct at both sites: with
`RUST_CRATE_SOURCE_ROOT_DIR_NAMES` renamed to non-existent names, a planted
`println!` in `sdk/streamlib-error/src/lib.rs` and a planted
`SchemaIdent::new(Org::new(...))` in `examples/hello-streamlib/src/main.rs` both
sailed through at exit 0 while `check-cdylib-reach` exited 1. Both now track
per-root counts, refuse per root naming the empty one, and carry
`check-cdylib-reach`'s test shape pinning every root against the real workspace.
`lint-logging` also gains `packages` in its opted-in-crate zone list — the only
tree holding `processors/`-rooted crates, so without it the gate declared a
source root nothing under it could ever hold (608 → 620 files, no new
violations).

**Re-verified after the rebase onto `main` @ 0.9.3:** `install-packages` 15/15,
the full `cargo xtask` gate battery (13 checks), and both `test.yml` legs —
`cargo test -p streamlib -p streamlib-macros -p streamlib-plugin-abi
-p streamlib-plugin-sdk -p streamlib-processor-extract --lib`,
`-p streamlib-engine --test attribute_macro_test`, and the plugin-ABI /
processor-extract `--tests` + `--doc` legs.


tato123:
## ⛔ DO NOT MERGE YET — two CI gates go silent under this branch's own guard

**(a)** fix the two gates before merging · **(b)** merge anyway and track the gate fix as a follow-up — reply **(a)** or **(b)**.

---

The loop exhausted all 6 fix rounds and escalated with this unresolved. It is not a style note; it was proven by a deliberate-break experiment, not by inspection.

This branch adds `ensure_source_walking_gate_read_source` (`xtask/src/main.rs:49`) whose own rustdoc says *"a gate whose scan roots moved out from under it is indistinguishable from a clean tree."* The guard does not work at 2 of its 3 new call sites.

**The experiment** — scan roots renamed to `["srcMOVED", "processorsMOVED"]` (exactly the scenario the guard names), with two genuine violations planted: a `println!` in `runtime/streamlib-engine/src/lib.rs` and a hand-rolled `SchemaIdent` literal in an example processor.

| gate | with broken roots | with roots restored |
|---|---|---|
| `lint-logging` | **exit 0** — "43 file(s) scanned, no violations" | exit 1 — catches the `println!` (608 files) |
| `check-processor-spec-new` | **exit 0** — "no hand-rolled SchemaIdent literals" | exit 1 — catches the literal |
| `check-cdylib-reach` | exit 1 — "scanned 0 files … the scan roots moved out from under the gate" | exit 1 |

So under this branch, **CLAUDE.md's first CI-enforced non-negotiable — no `println!` — can go completely unenforced with a green build.**

**Two distinct root causes:**

1. `lint_logging.rs:121-132` accumulates one `files_scanned` across `scan_target` (polyglot) and `scan_rust`. The 43 surviving Python/TS files mask the entire Rust half dropping to zero.
2. `check_processor_spec_new.rs:104-131` increments `files_scanned` for every `.rs` under `SCAN_DIR_PARENTS` regardless of `is_example_src_file`, so the count cannot notice that the root predicate stopped matching anything.

`check_cdylib_reach` is the only correctly wired caller — it records `scanned_file_paths` and pins per-root reach in `the_dispatch_scan_reaches_both_the_src_and_the_folder_backed_root` (`check_cdylib_reach.rs:1022-1038`). `check-no-escalate-in-lifecycle` was separately confirmed unaffected.

**Suggested fix:** wire both gates the way `check_cdylib_reach` already is — track per-source-root counts rather than one aggregate, and assert each configured root contributed at least one file. For `lint-logging` that means separating the Rust count from the polyglot count; for `check-processor-spec-new` it means counting files that satisfy `is_example_src_file`, not files read. Add the negative test in the same shape for both.

This is an instance of the class tracked in #1630 — a scan that finds nothing must fail loudly rather than report success.

tato123:
## Owner: (a) — 6 more fix rounds authorized, with a second item

Two things to resolve, not one.

### 1. The vacuous guard (as reported above)

Wire `lint-logging` and `check-processor-spec-new` the way `check_cdylib_reach` already is — per-source-root counts rather than one aggregate, asserting each configured root contributed at least one file, with the negative test in the same shape as `the_dispatch_scan_reaches_both_the_src_and_the_folder_backed_root`.

### 2. Per-platform re-export shims were supposed to disappear — they did not

Milestone #39's body, verbatim:

> Platform variants are two files declaring the same processor id under mutually-exclusive cfg; the id is the unified name, **so the per-platform re-export shims disappear**.

This branch moved `src/apple/` → `processors/apple/` wholesale and kept the shims. `packages/audio/processors/linux/mod.rs` is one:

```rust
#![cfg(target_os = "linux")]
pub mod audio_capture;
pub mod audio_output;
pub use audio_capture::{LinuxAudioCaptureProcessor, LinuxAudioInputDevice};
```

Nine packages carry this shape: `audio`, `camera`, `display`, `h264`, `h265`, `jpeg`, `mp4`, `screen-capture`, `webrtc`. The variants also still declare **distinct** processor ids (`LinuxAudioCaptureProcessor` / `AppleAudioCaptureProcessor`) where the design says the id is the unified name.

Nesting itself is fine and stays supported — `#[path]` in the generated root handles it, and the owner is explicit that folders are not the problem. What must go is the **re-export shim as the platform mechanism**: sibling files under `processors/` declaring the same id under mutually-exclusive `cfg`, with no `apple/mod.rs` or `linux/mod.rs` doing the routing.

`_apple_impl_pending_/` stays parked by `#[cfg(any())]` — that part is correct and must not change.

### Why this was missed

Thirty-five agents verified this branch and none checked it against the milestone's own architecture statement. The reviews are strong on empirical checks — the vacuous-guard finding was proven by deliberately breaking the scan roots and observing the gates go silent — and weak on "does this match the stated intent." Worth carrying into how the remaining tickets are reviewed.

Issue #1386, opened 2026-07-16
test(ci): audio_clock drop-counter test flakes under parallel execution (shared-global race)


tato123:
## Description

`core::plugin::host_services::audio_clock::…on_tick_drops_user_data_exactly_once_on_success_path` fails **intermittently under parallel `cargo test` execution** — a shared-global drop-counter race (observed failure: "before + 2" instead of the expected single drop). It passes single-threaded (`--test-threads=1`) and in isolation, so it's a test-harness race, not a product bug.

## Why it matters

`cargo test --lib` (the CI gate) runs tests in parallel by default, so this test can red a CI run that has nothing to do with the change under review — the same first-run-X churn we just eliminated for the cache step (#1378). It was surfaced (and proven unrelated) during the #1385 work: it flakes when the `audio_clock` module runs alone (run 1 FAILED, run 2 ok; no pubsub involved), and the full `host_services` suite passes single-threaded (309/0).

## Root cause (likely)

A drop-counter shared across test cases as a process-global `static` (AtomicUsize or similar) rather than per-test state; a concurrently-running sibling test increments/observes it, so the "exactly once" assertion sees another test's drops.

## Fix options

- Make the drop-counter **test-local** (per-test instance, not a module/process global), so parallel siblings can't perturb it. Preferred.
- Or gate the affected test(s) behind a `serial_test`-style serialization if a shared global is genuinely required.

## Exit criteria

- [ ] The test passes reliably under default parallel `cargo test --lib` (no shared-global cross-test contamination).
- [ ] No `--test-threads=1` requirement introduced for the suite.

## Related

- Surfaced during #1385 (`request_runtime_shutdown`); pre-existing, unrelated to that change.
- Same CI-flake-churn class as #1378 (Cache-Cargo). Low priority; mechanical.

tato123:
Closing — the test no longer exists. `core::plugin::host_services::audio_clock::…on_tick_drops_user_data_exactly_once_on_success_path` and the whole `core/plugin` tree went with the plugin-ABI ripout; `git grep` finds no trace. Duplicate of #1368's subject, closed for the same reason. Milestone sweep 2026-09-01.

Issue #1368, opened 2026-07-15
test(engine): fix the audio_clock on_tick parallel-run flaky test


tato123:
## Problem
`core::plugin::host_services::audio_clock::…on_tick_drops_user_data_exactly_once_on_success_path` intermittently FAILS under multi-threaded `cargo test --lib` (a shared process-global drop counter races); passes with `--test-threads=1`. Caught by #1253 + #1261 verify-change re-runs.

## Impact
Blocks gating `streamlib-engine`'s lib tests in CI — they'd red intermittently on this flake. Also intermittently reds any PR that does run the engine lib tests.

## Fix direction
Make the drop counter per-test isolated (thread-local / per-instance) or `#[serial]` that module. Then the engine lib tests can be added to the CI gate.

## Blocks
The engine-lib-tests-in-CI follow-up.

tato123:
Closing — the test no longer exists. `core::plugin::host_services::audio_clock::…on_tick_drops_user_data_exactly_once_on_success_path` and the whole `core/plugin` tree went with the plugin-ABI ripout (`[importable-python-library]`); `git grep` finds no trace. Nothing left to de-flake; the live engine-lib gating work is #1369 in Engine Test Suite Gated in CI (#47). Milestone sweep 2026-09-01.

PR #1656, opened 2026-07-29
fix(engine): compare port schema idents version-blind

tato123:
## Summary

- `classify_port_schema_agreement` and `classify_wire_schema_agreement` were the last two resolution surfaces still comparing a `SchemaIdent` version-inclusively, through its derived `PartialEq` over all four fields. Every other surface — registry lookup, loader cross-check, processor resolution, `ProcessorTypeReference` after #1648 — binds version-blind. Both now compare on the `(org, package, type)` identity tuple.
- Adds `SchemaIdentWire::matches_schema_tuple`, the wire-side mirror of the existing `SchemaIdent::matches_schema_tuple`, so the version-blind projection is named once rather than inlined into the classifier.
- The type-level check is untouched: a miswired `VideoFrame`→`AudioFrame` link still classifies as `Mismatch`, and #1430's revert locks stay green. Only the version axis is removed.

## Closes

Closes #1654

## Exit criteria

- [x] `@x/y/T@0.0.0` vs `@x/y/T@1.0.0` classifies `Compatible`, with a revert-lock test asserting it — `same_tuple_different_version_is_compatible`, plus the wire-side `wire_agreement_ignores_the_version_axis`.
- [x] `@x/y/VideoFrame` vs `@x/y/AudioFrame` still classifies `Mismatch`, existing revert locks green — `distinct_concrete_specs_mismatch` (widened to cover all three tuple segments, not just the type).
- [x] No spurious warn on a Rust→Python link — locked at both wiring sites, connect and read (see Test plan).

## What changed vs. the issue body

No divergence — every citation in the body verified accurate against current code. One imprecision worth noting for future readers: the table locates `classify_wire_schema_agreement` at `iceoryx2/input.rs:467-484`, which is its *caller*; the function itself is at `schema_agreement.rs:91-102`. The Design section cites correctly.

Per the body, the `Strict` posture is left in place — this is a comparator fix, not a policy change.

## Test plan

The issue notes this regression shipped twice because CI runs lib-tests only and never wires a live cross-language pipeline. Coverage was therefore added at the **wiring sites**, not just on the classifiers — this module's own doctrine (`operations_runtime.rs:641-645`) says classifier-only tests don't catch wiring-site regressions.

| gate | result |
|---|---|
| `cargo check` (both edited crates) | pass |
| `cargo test -p streamlib-ipc-types` | pass — 17 |
| `cargo test -p streamlib-idents` | pass — 256 |
| `cargo test -p streamlib-engine --lib` | 1742 pass, 2 pre-existing failures (below) |
| `cargo test -p streamlib --test app_sugar_test` | pass — 7 |
| `cargo doc` intra-doc links | clean |
| clippy on edited files | no new warnings; one pre-existing `if_same_then_else` removed |

**Revert-lock proof.** Rather than assert the locks bite, I reverted both comparators to `==` and re-ran: all five new tests fail, and every pre-existing mismatch test stays green — confirming the tests exercise the bug and that the type-level check was not weakened.

New tests:
- `same_tuple_different_version_is_compatible` / `wire_agreement_ignores_the_version_axis` — the classifiers.
- `loose_connect_is_silent_across_the_version_free_sentinel_asymmetry` / `strict_connect_accepts_the_version_free_sentinel_asymmetry` — through `connect_impl`, with a producer carrying the `0.0.0` sentinel and a consumer carrying `1.0.0`.
- `read_raw_is_silent_across_the_version_free_sentinel_asymmetry` — the read path, the other half of the spurious warn.
- `schema_ident_wire_matches_schema_tuple_ignores_version` — the new projection directly.

Both sources of the real asymmetry are already independently locked in CI, so the hardcoded pair in the connect test is provably representative: `grammar.rs:552,672` lock the `0.0.0` sentinel, and `project_config.rs:433` locks the manifest-inherited version.

### Live cross-language run — attempted, blocked

The Validation shape asks for a live cross-language link emitting no schema warning. I attempted it on `examples/polyglot-manual-source` (Python/Deno source → Rust sink, no camera or display). It confirmed the asymmetry on real in-tree code — `streamlib link ./python` and `./deno` both report `frame_out (@tatolab/core/VideoFrame@1.0.0)`, while the Rust sink declares the version-free `input("video_in", "@tatolab/core/VideoFrame")` → `@0.0.0`.

The runtime connect was not reached: the counting-sink plugin needs `_generated_rust_crate_root_/lib.rs`, which `cargo xtask generate-crate-roots` only produces for in-tree packages and the build orchestrator only produces during a real package install, not a `link`. `setup.sh` + `streamlib install` therefore do not leave that example runnable on unmodified `main`. Pre-existing gap, unrelated to this change; not patched here. The environment was fully restored afterward (`streamlib unlink --engine`, generated artifacts removed).

## Pre-existing failures (not introduced here, not fixed here)

Both fail identically on clean `main` at `731585ad`:

- `processor_spec::tests::serde_emits_structured_name_object_not_joined_string` — expects `"1.0.0"`, gets `Null`.
- `module_loader::lazy_load_error_tests::lazy_module_load_failed_names_type_package_and_detail` — expects `SemVer 1.0.0`, gets `0.0.0`.

Both are #1648 fallout: tests still asserting a version on a now-version-free *processor* reference. Same regression class as this ticket, on the processor axis rather than the schema axis.

Also observed: `module_loader::from_source::tests::stages_python_source_into_a_loadable_session_package` is flaky — failed once under the parallel suite, passes in isolation and on re-run.

## Notes

- **Derived-`PartialEq` trap.** The issue records that #1460 fixed this class and #1477 reintroduced it three weeks later *through derived `PartialEq`*. `SchemaIdent` and `SchemaIdentWire` both still derive version-inclusive equality (correct for byte-identity, wrong for agreement), so each now carries a note pointing agreement comparisons at `matches_schema_tuple`. Cheapest available guard against a third recurrence.
- **Extraction, called out per CLAUDE.md.** `compiler_with_pair` is extracted from `compiler_with_mismatched_pair` so the mismatched and cross-language shapes share one node-construction body — real duplication, not a speculative refactor.
- **Polyglot sweep: no migration needed.** No Python/Deno schema-agreement classifier exists; the SDKs only construct and encode `SchemaIdentWire`. Deno's `SchemaIdent.equals()` and Python's frozen-dataclass `__eq__` are version-inclusive but have no agreement callers, so there is no old pattern to migrate. Adding version-blind helpers with zero callers would be speculative API.
- This removes the constraint #1651 currently has to legislate by review convention — the `0.0.0` sentinel no longer has to be mandated on every polyglot port `SchemaIdent`.

## Follow-ups

None filed. Two candidates for the owner's call: the two #1648-fallout test failures above, and the `polyglot-manual-source` linked-package provisioning gap.

### DIFF
--- a/runtime/streamlib-engine/src/core/runtime/operations_runtime.rs
+++ b/runtime/streamlib-engine/src/core/runtime/operations_runtime.rs
@@ -667,6 +667,8 @@ mod connect_schema_agreement_tests {
 
     const PRODUCER_TYPE: &str = "SchemaMismatchProducer";
     const CONSUMER_TYPE: &str = "SchemaMismatchConsumer";
+    const SENTINEL_PRODUCER_TYPE: &str = "VersionSentinelProducer";
+    const VERSIONED_CONSUMER_TYPE: &str = "VersionedSchemaConsumer";
 
     fn ident(package: &str, ty: &str) -> SchemaIdent {
         SchemaIdent::new(
@@ -679,6 +681,15 @@ mod connect_schema_agreement_tests {
 
     fn schema(ty: &str) -> PortSchemaSpec {
         PortSchemaSpec::Specific(ident("core", ty))
+    }
+
+    fn schema_at(ty: &str, major: u32, minor: u32, patch: u32) -> PortSchemaSpec {
+        PortSchemaSpec::Specific(SchemaIdent::new(
+            Org::new("test").unwrap(),
+            Package::new("core").unwrap(),
+            TypeName::new(ty).unwrap(),
+            SemVer::new(major, minor, patch),
+        ))
     }
 
     /// Register a producer type (`out` → VideoFrame) and a consumer type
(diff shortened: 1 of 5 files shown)


tato123:
### Reviewer note: these tests are not CI-gated

Flagging so the green check table above isn't read as stronger coverage than it is.

None of the five tests added here run in CI. `.github/workflows/test.yml:79` runs `--lib` for `streamlib`, `streamlib-macros`, `streamlib-plugin-abi`, `streamlib-plugin-sdk`, and `streamlib-processor-extract`. The new tests live in `streamlib-engine` (4) and `streamlib-ipc-types` (1) — neither package is in that list, and of the engine CI runs only `--test attribute_macro_test`.

That exclusion is deliberate and tracked: #1369 gates `streamlib-engine`'s lib tests, pending the parallel-run flake fix in #1368 / #1386.

Practical consequence: the revert locks are real — I verified each one by reverting both comparators to `==` and confirming all five fail, then restoring — but they only fire when someone runs them locally. They do not defend the branch on their own until #1369 lands.

This sharpens the issue body's diagnosis, which said this regression "never surfaced because CI runs lib-tests only." The more precise version is that CI does not run the engine's lib tests at all — which is why the same defect could ship twice (#1460 → #1477) without anything going red.

Nothing here argues against merging; the fix and its locks are correct either way. It just means #1369 is what converts these locks into an actual gate.

Two related items filed from this branch's baseline run:
- #1658 — two engine lib tests still asserting a version on a version-free processor reference (#1648 fallout). Pre-existing on `main`, invisible to CI for the same reason.
- #1657 — the linked-package provisioning gap that blocked the live cross-language verification this issue's Validation shape asked for.

PR #1703, opened 2026-08-01
chore(spike): Tier A harness for the in-process Python go/no-go (#1702)

tato123:
## Summary

Tier A of the #1702 go/no-go spike: a throwaway crate that measures in-process Python (CPython
embedded via PyO3) against today's subprocess-per-Python-processor model, plus proof it runs end
to end. **Zero engine changes** — every processor enters through the existing `App::add_local`
path, and nothing outside `spikes/` is touched.

**This PR delivers the harness, not the protocol numbers.** No 10-minute cells, no full A/B/A
matrix, no soak, no subprocess baseline. Those are PR 2 (stacked: `spike/1702-pyembed-tier-b`).

### The interesting early result

| cell (10s smoke, not a protocol run) | emit→sink p50 | stage callback p50 |
|---|---|---|
| 720p30 Rust floor (no interpreter) | 9.63 ms | — |
| 720p30 Python passthrough | 9.61 ms | 13.8 µs |
| 1080p30 Python realistic | 27.16 ms | 4.18 ms |

The Python arm is **indistinguishable from the pure-Rust floor**. The engine wire hop outweighs the
PyO3 callback by roughly three orders of magnitude. That attribution is only possible because of
the floor arm, which was added after recon showed gate 5 passes by ~39x naive / ~3300x anchored and
therefore does not discriminate.

### Owner decisions that shaped this, recorded on the issue

Two `[NEEDS DECISION]` comments were posted before and during implementation
([first](https://github.com/tatolab/streamlib/issues/1702#issuecomment-5153081295),
[second](https://github.com/tatolab/streamlib/issues/1702#issuecomment-5153235020)). Answered:
`every_sample` delivery profile with latency percentiles primary; metric renamed to
`source_emit_to_sink_receive` because capture-to-present is not observable in either tier; floor
arm added and gate 5 demoted; Tier B goes 720p30 + 720p60.

**Still open and blocking the protocol runs, not this PR:** SCHED_OTHER vs SCHED_FIFO for the
primary numbers; whether to add an absolute p99.9 ceiling; whether 1080p60 is dropped (see below);
and whether the embed-vs-import arrangement needs measuring both ways.

## Closes

Closes nothing — #1702 stays open until the verdict is posted after Tier B.

## Exit criteria

Against #1702's "Done means", this PR covers item 2 (the artifact machinery) and none of item 1 or 3:

- [x] Artifact dir emitted per cell: runner script with `--fps --stage --mode --gc --duration`,
      `machine-spec.json`, per-frame JSONL, mergeable HDR histogram export, `summary.json`,
      GC telemetry sidecar, repro README.
- [x] Timer overhead microbenched and the stamping-compiled-out control implemented and proven.
- [ ] Tier A and Tier B numbers under the frozen protocol — PR 2.
- [ ] A verdict posted — PR 2.

## Test plan

- `cargo test --release` → **63 pass**, and `--features stamping-compiled-out` → **47 pass**
  (the recorder's tests are correctly absent from a control build).
- `cargo clippy --all-targets` → **clean** for this crate.
- `python3 -m unittest discover -s python` → **14 pass**, including one that invokes the harness
  with the exact argv `runner.py` builds and asserts a `summary.json` appears.
- End-to-end verified on the rig for both arms at 640x480 / 720p / 1080p, 30 and 60fps, GIL anchor
  on and off, and through `runner.py` with A/B/A interleaving. Zero drops, zero negative-latency
  anomalies, zero histogram saturation across every run.
- Load-bearing claims are asserted, not commented: zero-copy is proven by Python writing through
  into the Rust buffer and by invocation cost being flat across a 6.75x frame-size change; the GIL
  anchor is proven by a measured 4.1x end-to-end difference against its own control condition.

## Notes for owner

**1. The engine wire hop is the dominant cost, and it bounds the protocol.** Floor-arm p50 is
1.44ms at 640x480, 9.63ms at 720p, ~27ms at 1080p — pure Rust, no Python. `read_raw` allocates a
fresh 64 KiB `Vec`, then a fresh full-size `Vec`, then memcpys, **per read per hop**
(`plugin-abi/src/vtables/input_mailboxes.rs:156-157`, allocation inside the retry loop). At 1080p
the two-hop service time exceeds the 60fps frame period, so a 1080p60 cell runs saturated and its
percentiles would describe queue occupancy rather than latency. `read_raw_bounded` exists on the
engine's `InputMailboxesInner` (`iceoryx2/input.rs:448`) and would allow a reused buffer, but is
not exposed on the SDK's input reader. **This is a finding about the engine that is independent of
the pivot** — worth your call whether it becomes real work.

**2. The spike's `Cargo.lock` is not committed**, following `.gitignore:51-52` rather than
overriding it. For a crate whose output is gating numbers, caret-ranged deps mean a rerun can
resolve different versions. Flagged rather than decided.

**3. Nothing here is an API proposal.** The callback token registry, the process-global measurement
collection point, and the Python shapes are deliberate spike artifacts. #1702 defers API design
until after the numbers.

**4. Two reviewers ran against this branch and both initially returned REJECT.** Every blocker and
should-fix is addressed in the second commit; the most serious were that `runner.py` could not
invoke the harness at all, and that the `stamping-compiled-out` control feature had no `cfg` sites
and would have reported zero instrument overhead by construction.

**5. Known gap carried to PR 2:** escape detection via `Py_REFCNT` is exact only on a GIL-based
CPython; on a free-threaded build (3.13t+) biased reference counting makes it advisory. Also, a
callback that stashes `frame_view.ctypes.data` as a raw integer escapes the address without holding
a reference and reads as `NoEscape`.

### DIFF
--- a/spikes/streamlib-pyembed-spike/.gitignore
+++ b/spikes/streamlib-pyembed-spike/.gitignore
@@ -1,0 +1,2 @@
+/target
+/artifacts
--- a/spikes/streamlib-pyembed-spike/Cargo.toml
+++ b/spikes/streamlib-pyembed-spike/Cargo.toml
@@ -19,19 +19,16 @@ publish = false
 
 [lib]
 name = "streamlib_pyembed_spike"
-# `rlib` feeds the Rust-launched Tier A harness binary; `cdylib` is what CPython
-# dlopens when the spike is driven from `python3 python/runner.py`. The two are
-# built in separate invocations because they need opposite `extension-module`
-# settings — see the feature comment below.
-crate-type = ["rlib", "cdylib"]
+# `rlib` only. A `cdylib` + `extension-module` pair would claim CPython can
+# import this crate, but there is no `#[pymodule]` here: Tier A embeds CPython in
+# a Rust binary instead. #1702's design sketches the opposite arrangement
+# (a `PyApp` `#[pyclass]` CPython imports), and which one the pivot ships is an
+# owner decision — see the embed-vs-import note posted to the issue. Declaring an
+# unusable build target would only make that open question look settled.
+crate-type = ["rlib"]
 
 [features]
 default = []
-# PyO3's `extension-module` omits the libpython link. It MUST be on when CPython
-# dlopens our cdylib (the interpreter already provides the symbols) and MUST be
-# off when a Rust binary embeds the interpreter (nothing else supplies them).
-# The Tier A harness binary is the embedding case, so this stays off by default.
-extension-module = ["pyo3/extension-module"]
 # Compiles every measurement stamp out. The control cell for "does the
 # instrument move the measurement" builds with this on; a throughput delta
 # between it and a default build is instrument overhead, not signal.
@@ -55,6 +52,9 @@ serde = { version = "1", features = ["derive"] }
 serde_json = "1"
 clap = { version = "4", features = ["derive"] }
 tracing = "0.1.41"
+# Used only for the post-stop scoped subscriber in the harness binary: the
+# engine owns the global dispatcher and drops it at `Runner::stop()`, so a cell's
+# completion and invalidation records need a subscriber of our own to reach a log.
 tracing-subscriber = { version = "0.3", features = ["env-filter"] }
 
 [lints]
--- a/spikes/streamlib-pyembed-spike/Cargo.toml
+++ b/spikes/streamlib-pyembed-spike/Cargo.toml
@@ -1,0 +1,72 @@
+# Copyright (c) 2025 Jonathan Fontanez
+# SPDX-License-Identifier: BUSL-1.1
+
+# Throwaway spike for #1702 — measures in-process Python (CPython embedded via
+# PyO3) against today's subprocess-per-Python-processor model. Not a workspace
+# member and not a distributable: it self-roots via the `[workspace]` table
+# below, the same convention `examples/*` and most `packages/*` follow (see the
+# root Cargo.toml comment at :44-56). Nothing in the engine depends on it.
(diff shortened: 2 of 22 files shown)


tato123:
Superseded by the payload-size sweep recorded on #1702 ([comment](https://github.com/tatolab/streamlib/issues/1702#issuecomment-5153400372)):

**Note 1 in this PR's description overstates the problem.** The ~27ms floor at 1080p is an artifact of this branch's synthetic transport pushing whole uncompressed pictures through main memory, not a property of the engine's normal path. `VideoFrame` already references a GPU surface by id (`packages/core/schemas/video_frame.yaml:29`), so a real frame on the wire is a few hundred bytes.

Measured on the rig, in-process Python arm, 60fps, three-stage graph:

| payload | emit→sink p50 |
|---|---|
| 16 KB | **0.100 ms** |
| 300 KB | 0.312 ms |
| 3.6 MB | 6.799 ms |

At reference-sized payloads the whole pipeline including the Python callback is 0.1ms — about 0.6% of a 60fps frame budget. **The 1080p60 saturation concern is withdrawn**; it was self-inflicted by the spike's transport choice.

What still stands from note 1: the engine's guess-small-then-reallocate pattern (`plugin-abi/src/vtables/input_mailboxes.rs:156-157`) is a real inefficiency that every in-process stage pays whenever data does travel through main memory. Owner decision: recorded as a finding, not fixed here, worth its own ticket.

The code in this PR is unchanged and still correct — the synthetic source's payload shape becomes reference-sized in the Tier B branch, where it is a two-line change to the source's configuration.

Issue #1223, opened 2026-06-07
compiler: no-op sink for declared output ports with no downstream link


tato123:
## What we want

A declared output port with no downstream link is a legal idle state, symmetric with an
unconnected input. A producer stands up before its consumer exists without erroring at frame
rate.

## What we do today

`OutputWriter::write_raw` looks the port up in `self.channels` and errors if it is absent
(`runtime/streamlib-engine/src/iceoryx2/output.rs:272`):

```rust
let egress = channels
    .get_mut(port)
    .ok_or_else(|| Error::Link(format!("Unknown output port: {}", port)))?;
```

The map is keyed on links that were actually opened, so a declared-but-unlinked output is
simply absent from it. There is no no-op path. Verified against the tree 2026-09-06 — the
symptom is live.

The asymmetry is the defect: an unconnected **input** is a clean idle (the consumer never
fires), while an unconnected **output** is a per-tick error.

## Why we're moving away from the original body

The symptom held; the mechanism section did not. It was written against the plugin era and
names things the tree no longer has: `OutputWriterInner.connections` (now `self.channels`),
`add_connection` and the compiler op `open_iceoryx2_service_op`, "engine source v0.4.36", and a
repro built on `@tatolab/jpeg`'s `JpegDecoder` — a plugin package deleted with the plugin ABI
(#1715). The drone-racer runner it was hit in is not in the tree. Re-derive the repro from the
engine, not from that graph.

## Constraints that are facts

- A genuinely **undeclared** port name must still error. Only a declared port with zero links
  becomes a silent no-op — that distinction is the whole ticket.
- Plan §Processor model & scheduling: a port that loses a bag counts it. A no-op sink writes to
  nothing, so it must not be confused with a drop — it is not a loss, and it must not appear as
  one in `graph`.

## Skills that need to run, and who runs them

`/implement` — this is engine work under shipped behavior, not a plan decision.

## Debate it on evidence

An engine test standing up a processor with a declared, unlinked output and running it for
several ticks without error.

<details>
<summary>Original body (preserved by tracker reconciliation 2026-09-06)</summary>

## Problem

A processor that writes to a **declared** output port which has **no downstream connection** errors on every `process()` tick:

```
process() failed: OutputWriter::write_raw(port='video_out') failed: Link error: Unknown output port: video_out
```

[...]

PR #1720, opened 2026-08-03
feat(python)!: in-process authoring — capability-typed contexts, @input/@output, rt.add, GIL-release contract

tato123:
## Summary

A Python class decorated with `@processor` now runs in-process on its own engine thread, exchanging bags with other Python processors in the app's own interpreter.

```python
@processor
class BrightnessFilter:
    frames_from_upstream = LinkInputDataPort()
    frames_to_downstream = LinkOutputDataPort()

    def __init__(self, gain: float = 1.0) -> None:
        self.gain = gain

    def process(self) -> None:
        frame = self.frames_from_upstream.read()
        if frame is not None:
            self.frames_to_downstream.write({"value": frame["value"] * self.gain})
```

Registration rides `register_dynamic` — the one registration path that survives the plugin-ABI deletion in Change B — and the PyO3 host lives in the wheel crate, so the engine stays Python-free. The GIL is attached for the span of a user callback and nothing else.

## Closes

Closes #1708

## Exit criteria

- `@processor` classes execute in-process on the dedicated-thread model, three execution modes, descriptor-driven priority ✅
- `rt.add(ImportedClass)`; the pipeline API is `add` / `connect` ✅
- GIL-release contract, with a test proving a blocked native call stalls no other Python processor ✅
- Pixels never cross into Python — a bag is a named map; the exchange surface is #1710 ✅
- Generated `.pyi` stubs ship in the wheel ⚠️ hand-written, not generated — see notes
- Single-processor test harness, no hardware ✅ `streamlib.testing.SingleProcessorTestPipeline`
- **Teardown re-proved against live processors** (your comment's added criterion) ✅ — #1707 could only run it on an empty graph

## Test plan

41 pytest (25 in the no-GPU CI scope, up from 20) + 6 Rust unit tests + stubtest + pyright, all green locally on a GPU rig.

Two gates were verified red as well as green, by reintroducing the defect:

| gate | with the defect | with the fix |
|---|---|---|
| `test_building_the_graph_from_two_threads_does_not_deadlock` | fails at 60s naming the last marker reached | < 1s |
| `mypy.stubtest` (stub entry deleted) | `Runtime.connect is not present in stub` | clean |

The GIL contract has one behavioural test, not two. A producer blocked inside `write()` under `lossless` backpressure completes ~50 writes in 3s instead of tens of thousands — genuinely parked in `write_raw` — while another Python processor manages 2436 ticks. An earlier interval-wait test was retired: it exercised engine code this branch does not touch, so no edit here could turn it red.

## Notes for owner

**1. The authoring grammar diverges from what I announced, and you should decide whether to keep it.** I announced a `ProcessorContext` pyclass with `ctx.read` / `ctx.write`. What I built declares ports as class attributes (`self.frames_from_upstream.read()`) and gives lifecycle hooks no arguments at all. It is better DX — a port is named once, and every call completes and type-checks — but it is a first-class public-API shape you did not approve, and your #1707 comment pointed the other way ("in-process the context is a PyO3 object — rewrite"). Rationale is recorded in `docs/decisions/importable-python-library.md`. Change A's MODIFIED entry still says `processor_context.py` carries over; if you keep this shape, that line needs superseding via `/align`, which this session cannot edit.

**2. `.pyi` is hand-written where the change file says "generated".** Agreed with you mid-session. PyO3's introspection is `experimental-inspect` and maturin's `--generate-stubs` RFC is unmerged, so hand-written is what every comparable PyO3 wheel does. `mypy.stubtest` covers the drift a generator would have prevented. Revisit condition is recorded in the new ADR.

**3. Two reviews returned REJECT on the same blocker, and they were right.** `add` / `connect` held the lifecycle mutex across a GIL release while `run()` / `shutdown()` take it holding the GIL — a wedge that takes the interpreter down rather than raising. Reproduced under gdb. Fixed, with the regression gate now running on every PR.

**4. Non-blocking findings, not ticketed:**
- Two `Runtime`s cannot run at once in a process (shutdown-signal ownership is process-global). Pre-existing; the test harness now refuses a nested pipeline with that reason instead of letting the engine's error surface.
- `clippy::result_large_err` fires 8× on this crate because the engine's `Error` is ≥168 bytes — engine-wide, inherited, not this branch's to fix.
- `spawn_deno_subprocess_op.rs` still refers to "the Python subprocess host", a concept this branch retires. Change B territory.
- The bag path allocates an intermediate `rmpv::Value` tree per frame. Fine while bags are small named maps; if they get hot, stream straight into the buffer with `rmp::encode`.
- A `print()` inside a processor still comes back as `stdio_interceptor — <text>` at WARN, and engine logging initializes once per process. Both pre-existing, both now sit under the primary authoring surface.

**5. One engine-layer line changed.** The engine logged every Python processor as a "Python subprocess host". Fixed where the label lives, per engine doctrine, rather than worked around in the wheel.

### DIFF
--- a/.github/workflows/python-wheel.yml
+++ b/.github/workflows/python-wheel.yml
@@ -10,6 +10,11 @@ jobs:
   wheel-interpreter-lifecycle:
     name: Wheel build + interpreter lifecycle (no GPU)
     runs-on: ubuntu-latest
+    # A lock-order defect between the lifecycle mutex and the GIL wedges the
+    # whole interpreter rather than raising, so its regression test can only go
+    # red by running out of time. Without a bound it would burn the default six
+    # hours instead.
+    timeout-minutes: 45
 
     steps:
       - uses: actions/checkout@v4
@@ -75,6 +80,11 @@ jobs:
       # hand-maintained stub safe. `pyright --verifytypes` cannot substitute:
       # it scores annotation completeness, so a stub describing a method the
       # binary dropped still reads as 100% complete.
+      # The crate's own unit tests need a linked interpreter, which this job
+      # already has. Without this step they run on developer machines only.
+      - name: Wheel crate unit tests
+        run: cargo test --locked -p streamlib-python-wheel --lib
+
       - name: Type stubs match the compiled module
         working-directory: sdk/streamlib-python-wheel
         run: .venv/bin/python -m mypy.stubtest streamlib._engine
@@ -100,6 +110,7 @@ jobs:
   python-types:
     name: Pyright (no build)
     runs-on: ubuntu-latest
+    timeout-minutes: 10
 
     # Deliberately independent of the wheel build. With `_engine.pyi` checked
     # in, pyright reads the stub rather than the compiled module — it never
--- a/.github/workflows/python-wheel.yml
+++ b/.github/workflows/python-wheel.yml
@@ -66,8 +66,18 @@ jobs:
         working-directory: sdk/streamlib-python-wheel
         run: |
           uv venv --python 3.12 .venv
-          VIRTUAL_ENV="$PWD/.venv" uv pip install pytest
+          VIRTUAL_ENV="$PWD/.venv" uv pip install pytest mypy
           VIRTUAL_ENV="$PWD/.venv" uvx maturin@1.9.6 develop
+
+      # The only check that compares the stub to the binary. It imports the
+      # built module, introspects it, and fails on a name or signature the
+      # hand-written `_engine.pyi` no longer describes — which is what makes a
+      # hand-maintained stub safe. `pyright --verifytypes` cannot substitute:
+      # it scores annotation completeness, so a stub describing a method the
+      # binary dropped still reads as 100% complete.
+      - name: Type stubs match the compiled module
+        working-directory: sdk/streamlib-python-wheel
+        run: .venv/bin/python -m mypy.stubtest streamlib._engine
 
       # The runner has no GPU. `Runner::start()` initializes a GPU context whose
       # DMA-BUF pool pre-warm needs a driver that can allocate exportable device
@@ -86,3 +96,31 @@ jobs:
         run: |
           mkdir -p "$XDG_RUNTIME_DIR"
           .venv/bin/python -m pytest tests/ -v -m "not requires_gpu"
+
+  python-types:
+    name: Pyright (no build)
+    runs-on: ubuntu-latest
+
+    # Deliberately independent of the wheel build. With `_engine.pyi` checked
+    # in, pyright reads the stub rather than the compiled module — it never
+    # imports anything — so this needs no Rust toolchain, no Vulkan, and no
+    # maturin, and finishes in seconds. It is also the check the author's editor
+    # runs: Pylance is pyright, so a green job here means working completion and
+    # correct hovers for anyone writing a processor.
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Install uv
+        uses: astral-sh/setup-uv@v5
+
+      # pytest only, so its `py.typed` annotations resolve in `tests/`. The
+      # package under test is read from source; nothing is installed.
+      - name: Prepare the environment pyright resolves against
+        working-directory: sdk/streamlib-python-wheel
+        run: |
+          uv venv --python 3.12 .venv
+          VIRTUAL_ENV="$PWD/.venv" uv pip install pytest
+
+      - name: Type-check the wheel's Python surface
+        working-directory: sdk/streamlib-python-wheel
+        run: uvx pyright@1.1.411
--- a/.github/workflows/python-wheel.yml
+++ b/.github/workflows/python-wheel.yml
@@ -29,7 +29,8 @@ jobs:
             pkg-config \
             protobuf-compiler \
             libvulkan-dev \
-            glslc
+            glslc \
+            python3-dev
 
       - name: Install jtd-codegen
         run: |
@@ -74,17 +75,19 @@ jobs:
           VIRTUAL_ENV="$PWD/.venv" uv pip install pytest mypy
           VIRTUAL_ENV="$PWD/.venv" uvx maturin@1.9.6 develop
 
+      # The crate's own unit tests link libpython rather than building as an
+      # extension module (the `extension-module` feature is off by default for
+      # exactly this), which is why `python3-dev` is in the apt list above.
+      # Without this step they run on developer machines only.
+      - name: Wheel crate unit tests
+        run: cargo test --locked -p streamlib-python-wheel --lib
+
       # The only check that compares the stub to the binary. It imports the
       # built module, introspects it, and fails on a name or signature the
       # hand-written `_engine.pyi` no longer describes — which is what makes a
       # hand-maintained stub safe. `pyright --verifytypes` cannot substitute:
       # it scores annotation completeness, so a stub describing a method the
       # binary dropped still reads as 100% complete.
-      # The crate's own unit tests need a linked interpreter, which this job
-      # already has. Without this step they run on developer machines only.
-      - name: Wheel crate unit tests
-        run: cargo test --locked -p streamlib-python-wheel --lib
-
       - name: Type stubs match the compiled module
         working-directory: sdk/streamlib-python-wheel
         run: .venv/bin/python -m mypy.stubtest streamlib._engine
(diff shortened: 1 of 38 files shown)


tato123:
## ⚠️ Major revision (2026-08-03, second pass): the old SDK's API is restored

The owner ruled the initially-shipped surface a regression — the pivot means "off the ABI, same host services and features," not a redesigned API. Commit a3a3bea9 restores parity, built by a five-agent analysis (typestate design, old-SDK inventory, host-services mapping, real-usage survey of `examples/` under an explicit owner override of the deny rule, grammar reconciliation) and a three-agent build (Rust writer, Python writer, adversarial reviewer).

### What an author writes now (back to the old shape, in-process)

```python
import streamlib
from streamlib import RuntimeContextFullAccess, RuntimeContextLimitedAccess, input, output, processor

@processor
class BrightnessFilter:
    @input(delivery_profile="every_sample")
    def video_in(self) -> None: ...

    @output()
    def video_out(self) -> None: ...

    def setup(self, ctx: RuntimeContextFullAccess) -> None:      # Full: gpu_full_access available
        self.gain = ctx.config.get("gain", 1.0)

    def process(self, ctx: RuntimeContextLimitedAccess) -> None:  # Limited: gpu_full_access AttributeError
        frame, stamp = ctx.inputs.read_with_timestamp("video_in")
        if frame is None:
            return
        ctx.outputs.write("video_out", frame, timestamp_ns=stamp)  # A/V-sync propagation
```

Restored end-to-end: **Full/Limited capability typestate** on every hook (the DECIDED phase axis; lease-guarded lifetime erasure, stash-safe where the engine sanctions it — `self._outputs = ctx.outputs` from setup works), `ctx.config`/`ctx.time`, `read_with_timestamp` + `write(timestamp_ns=)`, `monotonic_now_ns` (CLOCK_MONOTONIC, old contract) + `MonotonicTimer` (timerfd), `log.warn(msg, **attrs)` with attrs in the JSONL record and automatic processor attribution, `@input`/`@output` with advisory `schema=`, `SchemaIdent`, and GPU capability objects whose `acquire_pixel_buffer`/`acquire_texture`/`resolve_surface` hit the same host services the escalate ops used to reach — direct calls now. Pixel access (`as_numpy`/`__dlpack__`) raises `NotImplementedError` naming #1710, which now extends an existing handle.

### Adversarial review results

640k hammered stashed-ctx calls across run/shutdown from 4 threads: zero faults. It also found a real engine leak — `SurfaceStore::release` never evicted the cache clone `check_in` parks, so a released pixel buffer pinned its pool slot forever (fifth acquire of any shape failed; latent under the subprocess path too). Fixed at the engine layer with a regression test that acquires-and-closes past pool depth. Plus: unwind-safe lease revocation, `TFD_NONBLOCK` on the timer.

### Remaining owner flags (small, none blocking)
- Epoch duality: Python default stamps are CLOCK_MONOTONIC (old behavior, bug-compatible); Rust processors stamp MediaClock. Unifying is a plan decision.
- `ctx.config` returns a fresh dict per access (old SDK returned the stored dict) — safer, noted as a deviation.
- Stashed `gpu_limited_access` can still acquire after `shutdown()` (the owned Arc keeps the GPU context alive) — no crash; is silent post-shutdown allocation acceptable?
- `emit_python_processor_log_record` is `pub` on the engine's logging surface — reachable by any consumer, not just the wheel.

Gates: 72 pytest (GPU tier ×3) · 18 Rust unit tests · stubtest · pyright 0/0 · lint-logging · check-boundaries · engine diff = 40 added lines in 2 files.

tato123:
## Escalate door added + the complete sweep (ada065a0)

The owner called out that `escalate` was deferred with no owner — correct, and the full sweep proved it was a live footgun, not a future one.

### The evidence, from the complete sweep (every non-Apple Rust processor in `packages/`, every Python file in `examples/`)

**Rust ctx usage, all 30+ processors aggregated:** `gpu_limited_access` ×16 · `gpu_full_access` ×13 · `runtime_id` ×3 · `escalate` (+ resolve/acquire/audio_clock/runtime/processor_id) — and inside escalate closures, 22 distinct privileged ops (timelines, encoder sessions, readbacks, kernels, DMA-BUF imports, present targets). The canonical pattern is universal: **stash Limited in `setup` → worker thread → escalate once → per-frame on Limited** (`camera_linux.rs`; `display_linux.rs` additionally re-escalates on mid-stream color-format changes, degrading to drain mode on failure).

**Python examples, all 23 files:** `ctx.inputs`/`ctx.outputs`/`ctx.config`/`resolve_surface` as previously reported — plus the smoking gun the sample missed: `avatar_character.py:411` calls `ctx.gpu_full_access.acquire_surface(...)` under `# type: ignore[attr-defined]` with a TODO naming a future escalate op. Real code already needed privileged acquire outside a full-phase hook and reached past the type system to get it.

### What shipped

`gpu_limited_access.escalate(callback)` — the callback receives a temporary full-access capability and its return value passes through. Design point: it's the **same `GpuContextFullAccess` class** as `ctx.gpu_full_access`, over a second lease source — so the privileged op surface can never fork; #1710/kernel-session add methods, not mechanics. The callback's capability expires with the callback (revoked on the failure path too). Engine gate semantics identical to Rust: runtime-wide serialization, never nest, setup-shaped moments not per-frame.

Test is the camera pattern verbatim: stash in `setup`, escalate from a worker thread, acquire inside, assert the stashed escalated handle raises afterwards.

### Sweep residue (known, deliberate, named)
- `ctx.audio_clock` (2 Rust uses) — not exposed; audio built-ins stay native. Surfaces with a Python-audio ticket if one ever exists.
- `ctx.runtime()` (graph mutation from a processor, 1 Rust use) — deliberately absent from Python: code is the source of truth, the edit loop is `dev`.
- `cyberpunk_processor.py` is a third, older API generation (`ctx.input("x").get()`, `ctx.gpu.*`) matching neither the subprocess SDK nor this surface — not a parity target; its siblings (watermark/glitch/lower_third/avatar) are covered via the adapter surface that rides #1710.

Gates: 73 pytest (GPU tier) · stubtest · pyright 0/0 · 18 Rust unit tests · check-boundaries.

tato123:
## Known interaction, for the record (found while writing verified samples)

Declaring `schema=` on either end of a link currently breaks reads at runtime: the engine's read path hits the half-deleted schema registry (`schema '@tatolab/core/VideoFrame@1.0.0' … not in the runtime schema registry`). The declaration grammar itself works and is tested; the data-plane interaction resolves when **#1655** (delete the registry) and **#1679** (wire tag inert) land. Until then wheel authors should omit `schema=`.

Also reproduced #1223 from Python (declared-but-unconnected output port raises on write) — already tracked, and the one-bad-frame contract contained it as designed.

Both findings feed the upcoming schema-free-ports `/align` (owner-confirmed): the direction under discussion deletes what `schema=` fed rather than repairing it — TypedDict/annotations as the type story, `read(port, into=T)` as the strictness dial, pip as the cross-package registry.

tato123:
## Owner rulings on the four flagged questions (asked and answered in-session, 2026-08-03)

1. **Clock epochs → unify on CLOCK_MONOTONIC.** MediaClock's process-private epoch is the anomaly: it subtracts a per-process offset from the kernel clock, destroying cross-node comparability (needed by #1714 helper processes and multi-node) and diverging from V4L2/ALSA driver timestamps for nothing. Direction: MediaClock on Linux becomes raw CLOCK_MONOTONIC. This lands as its own decided change — the engine-wide near-zero-timestamp assumptions need an audit first — folded into the upcoming schema align or a small follow-on. Python's restored default stamp is already on the right clock; no change to this PR.

2. **ctx.config → fresh dict per access stands.** Safer than the old SDK's cached dict; mutating config was never sanctioned. Recorded as an intentional deviation.

3. **Post-shutdown acquire on a stashed gpu_limited_access → accepted as-is.** Matches Rust's owned-clone semantics; reachable only from a user-leaked thread; memory is freed when the Arc drops.

4. **emit_python_processor_log_record stays pub.** One logging pipeline; a stray Rust caller just emits a well-formed record. The doc line naming its purpose is the fence.

No code changes to this PR from any of these. The clock unification is the one follow-on and rides the align.

tato123:
**Amendment to ruling 1 (clocks), sharpened by the owner after posting:** the direction is one concept — the machine's monotonic clock — in every language. `MediaClock` **survives** as the cross-platform seam for reading it (its macOS impl is already machine-wide via `mach_absolute_time`); the fix is the Linux impl dropping its process-start offset to return raw `clock_gettime(CLOCK_MONOTONIC)`. Once the two clocks agree, the wheel's `media_clock_now_ns` duplicates `monotonic_now_ns` and is deleted — one exported time function. Audit-then-flip (9 call sites, 3 in `time_context.rs` with since-start semantics), riding the upcoming align, not this PR.

PR #1737, opened 2026-08-04
feat(python)!: zero-CPU-copy frame exchange — DLPack device tensors, numpy views, DMA-BUF

tato123:
## Summary

Zero-CPU-copy frame exchange for the Python wheel — the four seams #1708 stubbed, built to the owner ruling on #1710 (Option A, the staging blit, done in the shape #1714 mirrors).

**The user-facing result:**

```python
frame = VideoFrame.from_bag(bag)
with ctx.gpu_limited_access.resolve_surface(frame.surface_id) as surface:
    surface.lock(read_only=False)
    tensor = torch.from_dlpack(surface)   # CUDA tensor — zero CPU copies
    tensor[:] = model(tensor)             # GPU-side edit
    # exiting the block publishes the edit back to the surface
```

- **Host side**: `as_numpy()` / DLPack over the engine's own memory — skia/PIL/cv2 all eat the numpy view. Row pitch from the allocation, writable via DLPack v1.x negotiation, `device="cpu"` selects it explicitly.
- **Device side**: per-surface OPAQUE_FD staging + exportable timeline in the engine (`device_export_staging.rs`), blit sourced texture-first (VRAM→VRAM off the camera's ring, ~21 µs at 1080p; buffer→buffer PCIe fallback), source resolved **per refill** so rotating producers are never stale. `__dlpack_device__`/`__dlpack__` decide the side once per handle and never disagree.
- **DMA-BUF export/import** for native code that speaks that dialect.
- **Pooled textures** registered at acquire under a minted id — the same staging blit serves them (ruling's in-scope item).
- **Lifetime**: engine value, staging, and CUDA import all ride the capsule owner — a tensor outliving its handle keeps live memory across all three allocators; pool-slot reuse is gated on release.
- `acquire_device_exchange_buffer` (an earlier inline design) does not exist; the plan gate caught it and the ruling replaced it.

Two independent review rounds (correctness + Rust craftsmanship) ran before this PR; every blocker and should-fix is addressed or listed below. The correctness reviewer live-reproduced a wrong-pixels bug on the camera (staging source frozen at creation → device tensors lagged a ring cycle); the fix is per-refill resolution, locked by a camera test asserting device==host pixels across 12 frames.

## Closes

Closes #1710

## Exit criteria

- [x] Frames/textures expose `__dlpack__` over the Vulkan↔CUDA path (OPAQUE_FD → external memory → device pointer) — graph frames and pooled textures, verified with real `torch.from_dlpack` on the rig
- [x] DMA-BUF export/import round-trips
- [x] CPU numpy fallback (via the host mapping — ticket body's cpu-readback-adapter line superseded by recon, see ticket)
- [x] Sync/lifetime/layout owned by the engine's export surface; ring-slot reuse gated on consume
- [x] Regression tests: use-after-free (tensor outlives frame/handle, both allocent layers), stale-source rotation (camera), write-publication (explicit unlock and with-block spellings)
- [x] Demo: processor mutates synthetic frames in place via numpy and via torch; re-resolve proves in-place

## Test plan

- `pytest tests/` (wheel venv): **100 passed** — incl. camera rotation identity, with-block publish, pooled-texture device export, torch round-trip, three-layer lifetime
- `cargo test -p streamlib-python-wheel --lib` (26), `-p streamlib-adapter-cuda --lib` (34, incl. DLPack v0.8 + v1.x layout regression)
- stubtest + pyright: clean; `cargo xtask check-boundaries`: 7,444 files, no violations
- Rig: RTX 3090, driver 595.84, /dev/video0. **None of the Python evidence runs in CI** — both suites are `requires_gpu` and CI deselects that marker.

## Notes for owner

- **Latent bug (pre-existing, now nearer the surface):** `surface_store.rs:1005-1012` wraps checked-out fds unconditionally as `RhiExternalHandle::DmaBuf`, ignoring the `handle_type` the service returns. Unreachable today; #1714's OPAQUE_FD staging checkout would reach it.
- **Read-before-signal disposition:** the ticket's regression-test line assumed the lock waits. Publication is the engine's actual ordering point (a source waits on its own timeline before writing the surface id downstream); all wording now says so, and the camera identity test is the ordering evidence. A strict read-before-signal test would need a deliberately misbehaving producer — did not build one.
- **Helper placement:** in-process only here, by the ruling; the staging+timeline shape is what #1714 wires a child onto.
- **cudarc floor dropped to `cuda-12060`:** torch cu126 wheels bundle libcudart 12.6, and a 12.9-only symbol panics cudarc's loader when that copy wins the dlopen race. Every cudarc symbol used workspace-wide predates 12.0 (reviewer-enumerated).
- **Worth a learning doc:** the measured NVIDIA `compatibleHandleTypes` disjointness (OPAQUE_FD ⊥ DMA_BUF) — the engine records only the runtime symptom today.
- **Deferred polish** (reviewer lows): `surface_format_name` could be `&'static str`; `bind_cuda_device_by_uuid` leaves the thread's CUDA device changed (documented behavior of import); `mod` ordering nit in `core/context/mod.rs`.
- **Pre-existing failures not touched:** `println!` clippy in three untouched adapter-cuda test files; ~117 fmt diffs in engine files outside this diff.
- **Darwin cross-check not runnable on this rig** (bzip2-sys cross CC missing). Non-Linux compilation addressed by reading (adapter-cuda's dlpack module now platform-agnostic; device paths cfg-gated) but not proven by a compiler.
- Frame-DX sugar (the `into=` cast protocol, `frame.gpu()`/`frame.cpu()`) is on #1731's agenda per owner approval — not in this diff.

## Summary by CodeRabbit

* **New Features**
  * Added GPU surface pixel exchange for Python through NumPy and DLPack, including zero-copy access, locking, row-pitch support, and device-write synchronization.
  * Added Linux DMA-BUF export and import for GPU surfaces and pixel buffers.
  * Added CUDA interoperability for sharing Vulkan-backed surface data without copying.
  * Added support for pooled textures and improved surface ownership and lifetime handling.
* **Compatibility**
  * Updated CUDA support to target CUDA 12.6.
  * DLPack functionality is available across platforms; CUDA-specific features remain Linux-only.
* **Documentation**
  * Documented NVIDIA external-memory handle compatibility and required cross-format transfers.

### DIFF
--- a/.github/workflows/python-wheel.yml
+++ b/.github/workflows/python-wheel.yml
@@ -73,7 +73,10 @@ jobs:
         working-directory: sdk/streamlib-python-wheel
         run: |
           uv venv --python 3.12 .venv
-          VIRTUAL_ENV="$PWD/.venv" uv pip install pytest mypy
+          # numpy rides along because the exchange-surface tests import it at
+          # module scope — collection needs it even when `requires_gpu`
+          # deselects every test that would use it.
+          VIRTUAL_ENV="$PWD/.venv" uv pip install pytest mypy numpy
           VIRTUAL_ENV="$PWD/.venv" uvx maturin@1.9.6 develop
 
       # The crate's own unit tests link libpython rather than building as an
@@ -134,7 +137,11 @@ jobs:
         working-directory: sdk/streamlib-python-wheel
         run: |
           uv venv --python 3.12 .venv
-          VIRTUAL_ENV="$PWD/.venv" uv pip install pytest
+          VIRTUAL_ENV="$PWD/.venv" uv pip install pytest numpy
+          # CPU torch: the exchange tests type-check against torch's API, and
+          # pyright reports a missing import as an error. The cpu wheel keeps
+          # the download tractable for a no-build job.
+          VIRTUAL_ENV="$PWD/.venv" uv pip install torch --index-url https://download.pytorch.org/whl/cpu
 
       - name: Type-check the wheel's Python surface
         working-directory: sdk/streamlib-python-wheel
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5520,6 +5520,7 @@ dependencies = [
  "rmpv",
  "serde_json",
  "streamlib",
+ "streamlib-adapter-cuda",
  "streamlib-media-builtins",
  "tracing",
 ]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5514,6 +5514,7 @@ dependencies = [
 name = "streamlib-python-wheel"
 version = "0.11.1"
 dependencies = [
+ "cudarc",
  "libc",
  "parking_lot",
  "pyo3",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -178,15 +178,15 @@ dlpark = { version = "=0.6.0", default-features = false }
 # `cudarc::runtime::sys::is_culib_present()` lets the adapter fail
 # gracefully on CUDA-less machines.
 #
-# Pin to `cuda-12090` so the build does NOT require `nvcc` (the
-# `cuda-version-from-build-system` alternative errors when nvcc is
-# absent). The `_v2`-suffixed external-resource symbols this binding
-# exposes remain ABI-stable through CUDA 13.x — runtime libcudart
-# continues to export them, so the dlopen path resolves regardless of
-# which CUDA version is installed on the runner. `default-features =
-# false` skips cuBLAS / cuRAND / NVRTC; we only need `runtime` (for
-# external-memory + external-semaphore APIs).
-cudarc = { version = "0.19", default-features = false, features = ["driver", "runtime", "fallback-dynamic-loading", "cuda-12090"] }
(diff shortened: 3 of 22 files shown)


tato123:
Note on the red "Wheel build + interpreter lifecycle" check (now rerunning): the failure is `core::runtime::runtime::tests::stopping_an_already_stopped_runtime_is_a_no_op` — "RuntimeStopping must be published exactly once across two stops" — a runtime shutdown-signals test this branch does not touch (zero overlap with the diff).

Evidence it is a pre-existing flake, not this PR:
- The **identical test with the identical assertion failed on main's own run** ([19:31 UTC today](https://github.com/tatolab/streamlib/actions), the run triggered by #1733's merge), and 3 of main's last 5 runs of this workflow failed.
- The exact CI invocation (`cargo test -p streamlib-engine --lib -- core::signals core::runtime::runtime::tests::stopping`) passes 3/3 locally on this branch.

This is a new flaky-test signature (not the tracked audio_clock flakes #1368/#1386). Not filing a ticket per the standing only-P0 rule — owner's call whether it joins the CI & Test Infrastructure milestone.

PR #1739, opened 2026-08-05
feat(cli)!: the dev experience — run/dev/new ship in the wheel's console script

tato123:
## Summary

`streamlib run` / `dev` move out of the Rust binary and into the wheel as its console script, which is the only host with an interpreter to execute `app.py` in. They resolve the entry file on the same convention the Rust verbs used, run it exactly as `python app.py` does (`__main__`, own directory leading `sys.path`, own `sys.argv`), call `setup(rt)`, and block in `rt.run()`.

A launched app hosts the one control plane, so it is as discoverable as any other host's node — the wheel takes the api-server path dep and `rt.host_control_plane()` is the opt-in. `streamlib new` scaffolds an app that works before the user has written anything. Wheels publish to each release with a PEP 503 simple index, so `pip install streamlib --index-url …` is one stable incantation.

**Breaking:** the Rust `streamlib run` / `dev` are deleted rather than left as a second launcher answering to the same name.

## Closes

Part of #1711 — deliberately not `Closes`. The remainder needs the effect running in a child
interpreter: the MVP-minute re-demo and a re-baselined frame floor (the current floor is
calibrated against in-process timing and will shift once each frame crosses an IPC hop). Both
are blocked by #1714, so this must not auto-close the ticket.

## Exit criteria

- [x] `dev`/`run` import `app.py` and execute `setup(rt)`; `-f <file>` overrides
- [x] `RunnerAutoBuild` off the run path
- [x] A bad save prints the app's own traceback and costs no engine boot
- [x] `streamlib new` scaffolds `app.py` + `pyproject.toml` + `.python-version` (3.12) + `.gitignore`, with the effect touching pixels through the exchange surface
- [x] The CLI ships as the wheel's console script (pip owns PATH)
- [x] Wheel release packaging + static PEP 503 simple index
- [x] CI green throughout; the Deno subprocess path still compiles
- [ ] ~~Dev-mode GIL-hold watchdog~~ — **removed on owner instruction**, see notes

## Test plan

Everything below was run on the rig (RTX 3090), not reasoned about.

- **136 → 134 wheel tests** (`pytest tests/`), GPU half included. New: 27 in `test_cli.py` (entry resolution parity with the six deleted Rust cases, `setup(rt)` execution, `__main__` semantics, app-relative imports, `sys.argv` isolation, traceback shaping, `SystemExit` passthrough, scaffold content), 6 in `test_cli_launch.py` (real nodes booted through the console script).
- **Red/green proofs**, not mental reverts: the scaffold frame-rate guard fails on the in-place effect (~25 frames/6s) and passes on the bulk-copy one; the entry-resolution and `sys.argv` tests were each checked against the reverted behaviour.
- `cargo test`: 30 wheel-crate, 91 CLI, 56 api-server.
- `xtask lint-logging` (654 files), `mypy.stubtest`, `pyright@1.1.411`, `cargo fmt` on every touched file — all clean. No new warnings in any file this PR touches.
- Two independent review rounds (`review-pr`, `rust-craftsmanship-reviewer`); a third was launched and killed as moot after the watchdog removal superseded its scope.

## Notes for owner

**1. The watchdog is gone, on your instruction.** #1711's Delivers listed a dev-mode GIL-hold watchdog. It was built, then removed: its premise was "one interpreter runs every Python processor", which is the model the architecture eliminates rather than reports. Two reviewers proved it also measured the wrong thing — it timed callback wall-clock, so a processor correctly releasing the GIL (`time.sleep`, numpy, `MonotonicTimer.wait()`'s 100ms default) got warned at and told to release the GIL it had already released. Ticket bullet is unmet by design.

**2. Two plan amendments are pending, and one gates merge in a reviewer's opinion.**
- §Processor model & scheduling still says *"Both placements are first-class: in-process (lowest latency, shares the app's interpreter)…"*. Owner position is that in-process is banned and `sys.executable` is the only mode. `/pivot` prompt drafted and handed over. **This text is what produced the watchdog** — it will keep producing drift until it changes.
- §Control plane still says *"`dev` binds loopback by default"*; `dev` binds `0.0.0.0` for mesh reach (owner call). `tests/test_cli.py` pins the behaviour and names the pending amendment in its docstring. A reviewer's position is that a test pinning code against an unamended DECIDED entry should not merge — your call.

**3. Engine-layer fix rode along.** An omitted `config` reached the engine as JSON `null`, and a built-in's config struct will not deserialize from `null` even though every field defaults. So `rt.add(CameraSource)` — the spelling ARCHITECTURE.md:143 blesses — died at graph-compile time, after `setup` returned and after every Python-side check passed. Now sends `{}`, matching `EmptyConfig`'s own stated wire convention. Fixed at the engine layer per the doctrine, not bandaided in the scaffold.

**4. Scaffold performance.** The first effect I wrote ran at ~4.4 fps in a release build: editing the mapped surface in place through a strided view costs ~225ms/frame, because the mapping is write-combined and CPU reads run ~175 MB/s. One bulk read, host-side edit, one bulk write back is 29.7ms — measured 7.6×. Worth knowing before anyone else writes a CPU-side effect; the exchange surface's `as_numpy()` docs say nothing about it.

**5. Observation verbs deferred.** `nodes` / `graph` / `tap` / `logs` / `mcp` are plan-promised for the wheel but still live only in the Rust binary, so a pip user can launch a node and not inspect it. Deferred with your agreement; the wheel now names where they live instead of argparse's `invalid choice`. Full porting inventory posted to #1712.

**6. Release workflow was unreachable as first written.** `on: release` cannot fire from a release release-please creates on the default `GITHUB_TOKEN`. Now a `workflow_call` invoked from `release-please.yml`, mirroring `trigger-schemas`. **Needs GitHub Pages enabled on this repo (source: GitHub Actions)** — a settings toggle I cannot make; until then the index publishes nowhere.

**7. Version sync.** `sdk/streamlib-python-wheel/pyproject.toml` was stale at 0.10.3 against a 0.12.0 workspace. Hand-synced once and added to release-please's `extra-files`, so the two artifacts cannot drift again.

**8. The scaffold is conformant; the engine still cannot enforce it.** `streamlib new` now writes the processor into `processors/inverting_effect.py`, so what it teaches matches the plan. But the `rt.add` wiring error §Processor model describes — refusing a `__main__:<Type>` identity with an error naming the fix — does not exist yet. A user who writes a processor into `app.py` today gets silence, not that error. That check is #1714's; the split makes the scaffold correct, it does not close the loop.

**9. Findings out of scope, not filed.** A native source with no downstream link logs `Unknown output port` every tick. `add_v_op.rs:63` stores `Some(spec.config)` unconditionally, so any front end still sending `Value::Null` hits the same failure as (3) — the engine does this itself at `open_iceoryx2_service_op.rs:1018`. First config error a user sees is `config deser: wrong msgpack marker FixStr(12)`, which the change file's "human-shaped failure messages" scope says it should not be. Pre-existing unused-import warning at `python_native_builtin_blocks.rs:11`.

## Summary by CodeRabbit

* **New Features**
  * Added Python CLI commands for creating, running, and developing Streamlib applications.
  * Added configurable control-plane hosting with host, port, and node name options.
  * Added camera and test-pattern project scaffolding.
  * Added automated wheel publishing and a pip-compatible package index.

* **Improvements**
  * Updated the Python package to version 0.12.0 and added the `streamlib` command.
  * Simplified handling of omitted processor configuration.
  * Moved application launch workflows from the Rust CLI to the Python CLI.


tato123:
**Placement-pivot hold (owner-approved tracker batch, 2026-08-05): do not merge as-is.** Two issues:

1. This PR carries `Closes #1711`, but #1711's remainder is now **blocked by #1714** — the scaffold and MVP-minute demo must be re-proven on helper placement (`[helper-process-placement-only]`, plan PR #1741). Merging would close the ticket with its remainder unbuilt.
2. The scaffold ships the banned shape: `cli.py` defines `@processor class InvertingEffect` inside the generated `app.py`, which `dev`/`run` execute as `__main__` — a `__main__`-defined processor class is now a wiring error at `rt.add` (reversal recorded in `docs/plan/changes/processor-class-identity.md`, plan PR #1741).

Either (a) strip `Closes #1711` and land the launch/packaging work as-is, with the scaffold split following behind #1714, or (b) move the effect class into its own importable module in this PR before merge. The watchdog removal already in this branch (`4caaa1fa`) is correct and stays.

tato123:
## Option (b) satisfied — ready to merge

Rebased onto `fdffb211` (all three pivots in), both requested changes done, everything green.

### 1. Scaffold split

`streamlib new` no longer puts a processor class in the entry file. §Processor model now rules that a wiring error:

> A processor defined in the entry file run as `python app.py` identifies as `__main__:<Type>` and is a wiring error at `rt.add` … The entry file itself may still run as `__main__`; only processor classes may not live there.

Layout is now `app.py` (docstring, one import, `setup(rt)`) + `processors/__init__.py` + `processors/inverting_effect.py`. The `process()` body is byte-identical, write-combined-mapping comment included.

`tests/test_cli.py:140` (`test_the_entry_runs_as_main_with_its_own_directory_importable`) is now load-bearing — it is what proves the `processors/` import resolves under `run_name="__main__"`. New `test_the_scaffolded_processor_lives_outside_the_entry_file` asserts the class is absent from `app.py` and present in the module, and parses both.

### 2. The release-wheel scaffold gate, shown red before green

The gate asserted only `test -f app.py`. Verified it green against a scaffold whose `processors/` directory was deleted outright — an app that raises `ModuleNotFoundError` on import. It now also requires the effect module, refuses a `@processor` in `app.py`, and imports the class by the exact name the engine will spawn it with.

One real bug surfaced only because of the red-first requirement: the entry-file check was first written `! grep -q '@processor' app.py`, and **`set -e` does not abort on a command whose status is inverted with `!`** — so that form stayed green on the violation it existed to catch. It is now an explicit `if … exit 1`.

| case | result |
|---|---|
| effect module deleted | exit 1 — red |
| class moved back into `app.py` | exit 1 — red |
| scaffold as shipped | exit 0 — green |

### 3. Linkage

`Closes #1711` → `Part of #1711`. The remainder — MVP-minute re-demo with the effect in a child interpreter, frame-floor re-baseline — is blocked by #1714 and must not auto-close. No auto-close keyword remains in the body.

### Verification

- 135 wheel tests, GPU half included — `test_the_scaffolded_app_reaches_a_running_graph` passes unchanged on the rig, as expected: the class still executes in-process until #1714.
- pyright 0 errors, `xtask lint-logging` clean across 653 files, `release-wheel.yml` parses.
- No watchdog or GIL diagnostic reintroduced; bind posture untouched. The bind test's docstring dropped its "amendment pending" note, since #1740 landed and §Control plane now states the all-interfaces default directly.

PR body note 8 records that the scaffold is now conformant while the `rt.add` wiring error itself is still #1714's to build — the split makes the scaffold correct, it does not close that loop.

PR #1751, opened 2026-08-05
refactor(adapters)!: survivor rewires A — python-native ABI edge cuts, surface-adapter carve-out, cross-process test re-homes

tato123:
## Summary

Survivor rewires A — the prep #1714 needs before Change B's crate deletions fire. Three deliverables; the fourth (the engine host-services strip) moved to #1715 on an owner ruling, with the reason recorded below and on both tickets.

**1. `streamlib-python-native` ABI dep edges cut.** Both `streamlib-plugin-abi` and `streamlib-adapter-abi` are gone from its `Cargo.toml`, so #1714 can port its Rust logic into the wheel before those crates are deleted.

Its only two plugin-ABI uses were `next_read_required_len` and `emit_channel_egress_admission_tracing` — iceoryx2 wire helpers, not ABI surface. Both are written entirely against `streamlib-ipc-types` vocabulary and exist so the host writer and the helper natives cannot drift from one `decide_channel_egress_admission` decision, so they now live beside that decision. `streamlib-ipc-types` gains `tracing`; `streamlib-plugin-abi` loses its `streamlib-ipc-types` edge (it had no other use). The Deno native and `iceoryx2/output.rs` follow the same move.

**One of the two changed shape rather than moving verbatim.** `next_read_required_len` computed `queue[queue.len() - 1]`, which underflows and panics on an empty slice, guarded by nothing but a doc sentence — a panic path in library code. It now returns `Option<usize>` over `first()`/`last()`. Both call sites already had a redundant `is_empty()` guard, which is deleted so the `let Some(..) else { continue }` is the single load-bearing emptiness check; semantics are bit-identical, since both arms continue to the next fan-in subscriber. The `Option` also distinguishes `Some(0)` — a zero-length frame in a non-empty queue — from `None`, which the `usize` return conflated.

**2. `adapters/streamlib-surface-adapter` — the surviving adapter cores' shared home** (name owner-chosen).

The split is by boundary, not by symbol list. The capability traits named in the change file (`VulkanWritable`, `GlWritable`, `CpuReadable`, …) are defined in terms of `ReadGuard` / `WriteGuard` / `StreamlibSurface`, so the whole non-cross-DSO half moves as one unit: `adapter`, `surface`, `error`, `guard`, `registry`, `mock`, `conformance`, `testing`, `subprocess_crash`.

`streamlib-adapter-abi` keeps exactly the cross-DSO half the plan retires with the plugin ABI — `ffi.rs` (`run_host_extern_c`) and `STREAMLIB_ADAPTER_ABI_VERSION`. Verified reachable only from the doomed `host_vtable.rs` halves, `core/plugin/host_services`, and `plugin-sdk`. The crate drops `bitflags`, `parking_lot`, `thiserror` and `libc` as a result.

**3. The spared cross-process GPU proofs re-homed** out of the `-helpers` crates the rip-out deletes, into the adapter cores they cover, where the `streamlib` dev-dep they need already lives:

| test | from | to |
| --- | --- | --- |
| `consumer_carve_out.rs` | `-vulkan-helpers` | `streamlib-adapter-vulkan/tests/` |
| `consumer_carve_out.rs` | `-cpu-readback-helpers` | `streamlib-adapter-cpu-readback/tests/` |
| `opaque_fd_consumer_rhi_round_trip.rs` | `-cuda-helpers` | `streamlib-adapter-cuda/tests/` |
| `opaque_fd_image_consumer_rhi_round_trip.rs` | `-cuda-helpers` | `streamlib-adapter-cuda/tests/` |
| `cuda_carve_out.rs` | `-cuda-helpers` | `streamlib-adapter-cuda/tests/` |

`streamlib-adapter-cuda-helpers` held nothing but those tests, so it is deleted here; its `cuda` feature and optional `cudarc` edge move onto `streamlib-adapter-cuda`, default-off exactly as before, and merge with that crate's existing (previously inert) `cuda` feature. The `-vulkan` and `-cpu-readback` helper crates stay — their subprocess-helper `[[bin]]` targets cannot see dev-deps, which is the whole reason those crates exist.

## Closes

Closes #1743

## Exit criteria

- [x] `streamlib-python-native` depends on neither `streamlib-plugin-abi` nor `streamlib-adapter-abi`
- [x] Adapter capability traits live outside `adapters/streamlib-adapter-abi`
- [x] All five spared cross-process GPU proofs survive outside the doomed `-helpers` crates
- [x] Workspace builds clean; `cargo fmt --all --check` has zero delta versus `origin/main`
- [x] The escalate GPU ops' dispatch path is untouched — #1714's privileged-access route is unaffected

## Test plan

- `cargo check --workspace --all-targets` — clean
- `cargo check -p streamlib-adapter-cuda --all-targets --features cuda` — clean (the `cuda`-gated carve-out test still compiles in its new home)
- Local gate battery via `local-ci-runner`
- `cargo fmt --all --check` diffed file-by-file against an `origin/main` worktree — zero files newly dirty (main carries ~92 pre-existing fmt-dirty files, mostly `vendor/`)
- All ten `cargo xtask` gates pass: `check-boundaries`, `check-consumer-rhi-repr`, `check-no-escalate-in-lifecycle`, `check-device-wait-idle`, `check-cdylib-reach`, `lint-logging`, `check-no-reverse-dns`, `check-no-inventory-submit`, `check-processor-spec-new`, `check-no-streamlib-metadata`; plus `check-vendored-vulkanalia` (`vendor/` and the license files are byte-identical to `main`)
- `cargo clippy --workspace --all-targets` — no errors (`disallowed_macros` is the only workspace deny); remaining output is pre-existing warnings
- `cargo test -p streamlib-ipc-types -p streamlib-plugin-abi -p streamlib-surface-adapter --all-targets` — 135 passed, 0 failed
- `cargo test -p streamlib-adapter-{cuda,vulkan,cpu-readback} --all-targets` — all pass

**All five re-homed proofs ran for real against the GPU on the rig in their new homes, and passed:**

| test | result |
| --- | --- |
| `-vulkan::consumer_carve_out` | 4 passed (concurrent submit serialization, leak-on-drop warning, silent drop, trait impl) |
| `-cpu-readback::consumer_carve_out` | `host_image_to_consumer_staging_byte_equal_round_trip` ok |
| `-cuda::opaque_fd_consumer_rhi_round_trip` | `opaque_fd_chain_host_export_to_consumer_import_to_adapter_acquire` ok |
| `-cuda::opaque_fd_image_consumer_rhi_round_trip` | `opaque_fd_image_carve_out_round_trip` ok |
| `-cuda::cuda_carve_out` (`--features cuda`) | `host_buffer_to_cuda_byte_equal_round_trip` ok — real `cudarc` import + byte-equal assertion |

They are unmodified apart from import paths, so their assertions are unchanged.

## Notes for owner

**The host-services strip is not a survivor rewire — it moved to #1715 (your ruling, 2026-08-05).** Recorded in full on #1743's body and as a comment on #1715. The short version:

- Change B's inventory says "~10 retained files that dispatch through `core/plugin/host_services`". The tree has **24 files and ~100 dual-mode sites** outside `core/plugin/`. The five heaviest — `vulkan_command_recorder.rs` (15), `vulkan_graphics_kernel.rs` (14), `vulkan_compute_kernel.rs` (14), `vulkan_ray_tracing_kernel.rs` (11), `core/context/gpu_context.rs` (11) — are not among Change B's eight named files. The 2026-08-02 audit under-counted ~3×.
- **The mode branch is also the cdylib safety guard.** Every PluginAbiObject's `host_inner()` reads `if host_callbacks().is_some() { panic!(…) } else { unsafe { &*(handle as *const Inner) } }`. Deleting the branch and keeping the host arm converts a caught panic into a raw deref of host-private layout — UB, not a diagnosable failure — while `plugin-abi` / `plugin-sdk` / the cdylib examples remain in-tree.
- Removing the `vtable` fields instead cascades: `core/plugin/host_services/gpu_context/limited/*` and ~40 siblings construct and read them, so the fields cannot go until `core/plugin/` goes — which is #1715.
- Nothing is green in between: 22 `load_project_dylib_*` / `pack_then_load_smoke` / `cdylib_owns_tokio_runtime` engine tests, 4 `export_plugin_*` plugin-abi tests, `twin_drift_guard` and `check-cdylib-reach.yml` all assert the behaviour the strip removes. All are already on #1715's deletion list.

**#1714 is unblocked without it.** The clause the ticket wanted the strip to protect — "the escalate GPU ops' dispatch path survives, it is the helper children's only privileged-access route" — is already true in the tree. `subprocess_escalate.rs`, `cpu_readback_bridge.rs`, `escalate_gate.rs` and `escalate_scope_registry.rs` carry **zero** references to `host_services` / `host_callbacks`. The needle is threaded as a verified finding rather than a diff.

**A plan edit is owed** (not editable from `/implement`): `docs/plan/changes/importable-python-library-ripout.md`'s host-services-strip bullet still carries the stale "~10 retained files" count and still sequences the strip ahead of the deletions it is actually inseparable from. Its "Adapter capability traits … shared home" bullet could also now name `streamlib-surface-adapter`, and its `REMOVED: streamlib-adapter-cuda-abi / streamlib-adapter-cuda-helpers` line is half-satisfied early by this PR.

**For #1714's awareness — the dep edge is cut, the dep *graph* edge is not.** Cutting `streamlib-python-native`'s two direct `Cargo.toml` entries does not remove `streamlib-adapter-abi` from its resolved dep graph: `cargo tree -p streamlib-python-native -i streamlib-adapter-abi` still shows it arriving transitively through all four adapter cores' surviving `host_vtable.rs` halves. The ticket's deliverable is worded "out of its Cargo.toml" and that is met — and the port #1714 needs works, because nothing in python-native names those crates any more — but the transitive edge only dies when #1715 deletes the cores' cross-DSO halves. Flagged so #1714 doesn't plan around a clean `cargo tree`.

**A clippy/lint-logging divergence worth recording.** `cargo clippy --locked --workspace --all-targets` fails `disallowed_macros` at four sites:

| site | macro |
| --- | --- |
| `adapters/streamlib-adapter-opengl/tests/bin/opengl_adapter_subprocess_helper.rs:41` | `eprintln!` |
| `adapters/streamlib-adapter-skia/tests/bin/skia_adapter_subprocess_helper.rs:55` | `eprintln!` |
| `adapters/streamlib-adapter-skia/tests/bin/skia_adapter_subprocess_helper.rs:185` | `eprintln!` |
| `runtime/streamlib-consumer-rhi/src/consumer_vulkan_device.rs:822` | `println!` |

All four lines exist verbatim on `origin/main`. Two of the three files are byte-identical to base; the skia helper differs only by this branch's `use streamlib_adapter_abi` → `use streamlib_surface_adapter` import block. So the branch neither caused nor worsened any of them, and they stay out of the diff per the doctrine on not auto-fixing unrelated findings.

The reason this isn't a CI failure: **`cargo clippy --workspace` is deliberately not a CI job** — `lint-logging.yml:14` records the rationale, and `cargo xtask lint-logging` is the gate that actually enforces the `println!`/`eprintln!` rule. It passes at 653 files, because both sites are test-only (a `tests/bin/` fatal handler and a `mod tests` skip message), which lint-logging exempts. So the two hits are a standing clippy-vs-gate divergence on `main`, not a regression here. Noting it so the next person doesn't re-discover it.

**Non-blocking observations, no tickets filed:**
- `sdk/streamlib-deno-native` also lost its `streamlib-plugin-abi` edge — unavoidable collateral of the helper move (it shared both functions), not scope creep. It dies with Change B anyway.
- The two re-homed `consumer_carve_out.rs` / `opaque_fd_consumer_rhi_round_trip.rs` files carried dead imports (`PixelFormat`, `ConsumerPixelFormat`) that had been warning in the helper crates; dropped in a separate commit since I was re-homing the files.
- `streamlib-adapter-cuda`'s pre-existing `cuda` feature gated nothing in `src/` — it had no `cudarc` dep and no `#[cfg(feature = "cuda")]` in the crate. Merging the helpers crate's feature into it makes the feature real for the first time.
- `adapters/streamlib-surface-adapter/Cargo.toml` is added to `check-boundaries`' `NO_STREAMLIB_RUNTIME_DEP` list. The old `streamlib-adapter-abi` occupied the same position and was never listed, so this closes a gap that pre-dates the branch — the new crate sits beneath all five adapter cores, where a `streamlib` dep would reach every one of their runtime graphs at once.
- Deleting `streamlib-adapter-cuda-helpers` *strengthens* the isolation invariant it existed to protect: `streamlib` and `streamlib-engine` now sit in `[target.'cfg(target_os = "linux")'.dev-dependencies]` of `streamlib-adapter-cuda`, which **is** in the `check-boundaries` list, whereas `-cuda-helpers` never was. `cargo tree -e normal` confirms no `streamlib` / `streamlib-engine` / `streamlib-sdk` in any of the five cores or the new crate, with or without `--features cuda`.
- `STREAMLIB_ADAPTER_ABI_VERSION` now versions a trait living in another crate and has no Rust consumer at all (only `docs/architecture/surface-adapter.md` mentions it). Left alone — the constant dies with its crate in #1715.
- **Same defect class as the `next_read_required_len` fix, ten lines below it, not fixed:** both natives do `let last = queue.len() - 1; let item = queue.swap_remove(last);` — the identical underflow shape. It is provably unreachable-as-a-panic (see below), it is untouched by this diff, and `queue.pop()` is **exactly equivalent** at the last index — a one-line, semantics-preserving swap for whoever picks it up. Reported rather than fixed, per the doctrine on unrelated findings.


tato123:
## Full local gate battery — result

Ran via `local-ci-runner` against every gate `.github/workflows/*.yml` actually runs, plus the xtask lint suite. **32 gates pass, 0 fail.** The run covered several I had not exercised myself:

- `install-packages` — 15/15 distributable packages compiled through the real user install path
- `check-manifest-schema` (115 `streamlib.yaml` validated), `check-pack-load`, `check-package-version-drift`, `check-processor-source-reachability` (97 files / 19 packages), `check-schema-versions`
- `xtask` fixture suite — 238 passed
- python-wheel chain: `maturin develop` builds, `stubtest` clean, `pytest` 83 passed, `pyright` 0 errors
- license-check header scan — 889 Rust files + all Python files OK (the local failures I saw earlier were `examples/*/target/` build artifacts absent from a fresh CI checkout)

**Two corrections to the runner's report, verified myself:**

1. It reported `cargo clippy --locked --workspace --all-targets` as **exit 0**. That is wrong — it fails `disallowed_macros` at four sites (listed in the PR description). All four lines exist verbatim on `origin/main`, so the conclusion "not caused by this branch" holds, but the gate is not green. `cargo clippy --workspace` is deliberately not a CI job (`lint-logging.yml:14` records why); `cargo xtask lint-logging` is the real gate and passes at 653 files.

2. The battery ran against `39f76708`, three commits behind the current HEAD `acfc61c4`. I re-ran the gates my later commits could affect — `cargo test -p xtask` (238 passed, covering the `check_boundaries.rs` edit), `check-boundaries` (7441 files, no violations), `lint-logging` (653 files, no violations).

**Broader non-gating `cargo test --workspace` fallback:** 201 test-result blocks ok, 5 failed. All five reproduce on `origin/main` and are unrelated to this diff — `git diff origin/main..HEAD` is empty for every implicated file:

| failing test | why it pre-dates this branch |
| --- | --- |
| `from_source::tests::stages_python_source_into_a_loadable_session_package`, `runtime::tests::stopping_an_already_stopped_runtime_is_a_no_op` | the known engine-lib parallel-run flake `test.yml` documents and excludes; both pass when run in isolation |
| `dlopen_camera_processor_completes_lifecycle_against_vivid` | shells to `cargo build -p streamlib-camera`; `packages/camera` is not a workspace member by design |
| `dlopen_cross_rustc_fixture_round_trips_every_plugin_abi_object` | fixture still uses a bare-string `#[processor("…")]`, made illegal by `1f6bda74` on `main`; never migrated |
| `failing_dlopen_load_leaves_zero_residue_and_reload_succeeds` | injects a manifest fragment with a `version:` field the schema rejects post-`1f6bda74` |
| `add_processor_with_unknown_type_returns_typed_error` | expects the old `9.9.9` sentinel; identity correctly returns `0.0.0` post-`1f6bda74` |

The last three look like real pre-existing breakage on `main` from the `1f6bda74` grammar change rather than flakes — worth knowing, though all three live in the cdylib/plugin corpus #1715 deletes.

Hardware-integration tier not run by the battery; the five re-homed cross-process GPU proofs were run separately on the rig and pass (see the PR description's test plan).

PR #1768, opened 2026-08-07
feat(engine): expose a graph-readiness signal the app process can poll after run()

tato123:
## Summary

`SingleProcessorTestPipeline.__enter__` returned as soon as it started the run loop, so a `feed` on the very next line published into a link whose consumer — the processor under test's helper process — attached tens of milliseconds later. The link drops what it carries in that window, so the bags were lost and the suite was skipped. A three-second settle was the only thing that made it pass.

The engine now has a real signal for that moment, and the harness blocks on it.

**Engine.** `StateComponent` carries a condvar beside the state, so a waiter blocks on the transition rather than sampling for it, and the ~8 open-coded `*state.0.lock() = X` sites go through `transition_to`. `Runner::observable_graph_readiness()` hands out the per-processor states themselves, so the wait runs with the graph lock released — the threads making those transitions need that lock — and without holding the engine.

The state attaches at `add_v` instead of at compiler-prepare time. That closes a false ready: a wait beginning before `start()` would otherwise find no states at all and report an empty graph as vacuously running. It is also what makes an early wait sound, since the compiler transitions that same state rather than replacing it.

**Wheel.** `run()` leaves a `Weak<Runner>` behind — weak because the teardown contract requires the run loop to hold the only strong reference — and `Runtime.wait_until_every_processor_is_running` reads the states in one detached step and waits in another, so the engine reference is gone before the long wait starts. The wait is legal before `run()` as well as during it: `Thread.start()` says nothing about when that thread reaches `run()`, and refusing the early call would leave the harness turning on the scheduling order it exists to remove.

The harness API is unchanged — the wait is invisible to the tests that use it.

## Closes

Closes #1759

## Exit criteria

- [x] A readiness observable the app process can reach after `run()`
- [x] `SingleProcessorTestPipeline.__enter__` blocks on it; no settle sleep anywhere
- [x] `tests/test_single_processor_pipeline.py` unskipped and green
- [x] The interpreter-lifecycle contract survives — teardown still joins every thread

## Test plan

- `tests/test_single_processor_pipeline.py` — 6/6, feeding on the first line of the `with` block. Five consecutive runs, then three full-suite runs under randomized ordering. The reviewer independently monkey-patched the readiness gate to a no-op and confirmed the scenario goes red (`produced nothing within 8.0s`), so the gate is proven rather than asserted.
- `runtime/streamlib-engine/tests/graph_readiness_signal_test.rs` (new, 4) — an empty graph is up by definition; a never-started graph is **not** reported as running (the false-ready guard); a failed processor ends the wait without burning the timeout; a failure sitting *behind* a processor that never starts still reaches the report.
- `state_component.rs` unit tests (4) — the wait ends on the transition and not on the deadline (a poll loop passes the state assertion and fails the timing bound); a failed setup ends it; an already-happened transition does not hang it; an expired deadline reports the state it gave up in.
- `tests/test_graph_readiness.py` (new, 7) — pre-run, post-shutdown, and four timeout values Python can express that a `Duration` cannot.
- `test_interpreter_lifecycle.py` — a readiness wait taken while `run()` blocks leaves teardown intact.
- Full wheel suite 174 passed / 1 skipped; GPU-free half 120 passed; `cargo test -p streamlib-engine --lib` 1771 passed.
- Gates: stubtest, pyright, `check-no-in-process-placement`, rustfmt on every changed file, clippy on both crates (no new hits).

## Notes for owner

**Two questions the reviewers raised as your call, not mine:**

1. `ObservableGraphReadiness` is new public surface on `streamlib::sdk::graph`, and the plan has no processor-state or readiness entry. The ticket body specifies "an accessor over the graph's `StateComponent`s", which this is, and it extends `StateComponent` rather than paralleling it — but does it earn a line in `ARCHITECTURE.md`, or is "ticket-level, no plan entry" the right answer?
2. `clone_inner()` fails the zero-context naming test on its own, but it matches the existing `ProcessorPauseGateComponent::clone_inner()` two lines away in the same function. Rename both, or grandfather it as an established component idiom?

**Deviation from the ticket.** The ticket sketched an accessor plus a Python-side poll loop. I built the signal instead (condvar, no polling), which left a bare `processor_states()` accessor with no callers — so it is not in the diff. The observable is `observable_graph_readiness()`.

**Found on the way, not fixed (outside this ticket):**

- `spawn_dedicated_thread` has four early returns between phases 3 and 4 (`spawn_processor_op.rs:283-322`) that log an error and abandon the thread *without* transitioning the state. Pre-existing and previously unwatched; now each one makes a readiness caller burn its full timeout and be told "had not finished setup (still Idle)" when the real cause was already logged. Cheap fix — transition to `Error` on each — but it is not what #1759 asked for.
- `spawn_python_native_subprocess_op.rs:216` waits for its child's `ready` with no deadline at all; the wheel host has a 60s one.
- `capture_helper_process_launch_environment` captures the app's import root **once per process**. The harness builds a fresh `Runtime` per pipeline, so two test modules in different directories both using the harness in one pytest process would hit this. This is what the CLI-suite `sys.path` leak in the third commit exposed.
- `GraphInspector` is dead code — no callers anywhere, and `core/observability/mod.rs` does not even re-export it.
- Pre-existing on `main`: unused `ProcessorTypeReference` import in `add_v_op.rs`; `cargo fmt --all --check` fails repo-wide on 204 files (not a CI gate, and the set includes `vendor/tatolab-vulkanalia*`, which is off-limits to reformat).

**Behaviour change worth knowing.** Moving `StateComponent` to `add_v` means a successfully-added, never-started node now renders `"state": "Pending"` in `GET /api/graph` where it previously had no state key. Additive, and arguably a fix — an un-started node now renders like a started one — but no test covers it. `to_json` also moved from `Debug` to `Display`; every variant produces an identical string, so no wire change.

## Summary by CodeRabbit

* **New Features**
  * Added runtime readiness checks that wait until all processors are running.
  * Added Python support with configurable timeouts and validation for invalid values.
  * Readiness failures now report processor states and startup errors.
* **Improvements**
  * Test pipelines wait for startup completion before accepting input, preventing initial data loss.
  * Processor lifecycle states are tracked consistently across startup, pause, resume, shutdown, and failures.
* **Bug Fixes**
  * Improved runtime teardown when readiness checks run during shutdown.
  * Improved diagnostics for invalid ports and startup failures.

### DIFF
--- a/runtime/streamlib-engine/src/core/compiler/compiler.rs
+++ b/runtime/streamlib-engine/src/core/compiler/compiler.rs
@@ -272,7 +272,7 @@ impl Compiler {
                     if let Some(node) = graph.traversal_mut().v(proc_id).first_mut() {
                         // Set state to stopping
                         if let Some(state) = node.get::<StateComponent>() {
-                            *state.0.lock() = ProcessorState::Stopping;
+                            state.transition_to(ProcessorState::Stopping);
                         }
                         // Send shutdown signal — eventfd wakes reactive
                         // mode's epoll, channel send wakes continuous/manual.
@@ -357,7 +357,7 @@ impl Compiler {
                     let mut graph = graph_arc.write();
                     if let Some(node) = graph.traversal_mut().v(proc_id).first_mut() {
                         if let Some(state) = node.get::<StateComponent>() {
-                            *state.0.lock() = ProcessorState::Stopped;
+                            state.transition_to(ProcessorState::Stopped);
                         }
                     }
                     // Remove from graph
--- a/runtime/streamlib-engine/src/core/compiler/compiler_ops/prepare_processor_op.rs
+++ b/runtime/streamlib-engine/src/core/compiler/compiler_ops/prepare_processor_op.rs
@@ -6,6 +6,7 @@ use crate::core::graph::{
     Graph, GraphNodeWithComponents, ProcessorPauseGateComponent, ProcessorReadyBarrierComponent,
     ProcessorReadyBarrierHandle, ProcessorUniqueId, ShutdownChannelComponent, StateComponent,
 };
+use crate::core::processors::ProcessorState;
 
 /// Attach infrastructure components to a processor node.
 /// Returns a barrier handle for coordinating with the processor thread.
@@ -25,8 +26,14 @@ pub(crate) fn prepare_processor(
     // Attach infrastructure components (NO ProcessorInstanceComponent - thread creates it)
     node_mut.insert(barrier_component);
     node_mut.insert(ShutdownChannelComponent::new());
-    node_mut.insert(StateComponent::default());
     node_mut.insert(ProcessorPauseGateComponent::new());
+
+    // Transitioned, never re-inserted: `add_v` attached the state when the node
+    // was added, and a fresh component would strand every waiter already
+    // holding the old one.
+    if let Some(state) = node_mut.get::<StateComponent>() {
+        state.transition_to(ProcessorState::Idle);
+    }
 
     tracing::debug!("[{}] Infrastructure components attached", proc_id);
     Ok(barrier_handle)
--- a/runtime/streamlib-engine/src/core/compiler/compiler_ops/spawn_processor_op.rs
+++ b/runtime/streamlib-engine/src/core/compiler/compiler_ops/spawn_processor_op.rs
@@ -17,9 +17,9 @@ use crate::core::descriptors::ProcessorRuntime;
 use crate::core::error::{Error, Result};
 use crate::core::execution::run_processor_loop;
 use crate::core::graph::{
-    Graph, GraphNodeWithComponents, ProcessorInstanceComponent, ProcessorPauseGateComponent,
-    ProcessorReadyBarrierComponent, ProcessorUniqueId, ShutdownChannelComponent, StateComponent,
-    SubprocessHandleComponent, ThreadHandleComponent,
+    Graph, GraphNodeWithComponents, ObservableProcessorState, ProcessorInstanceComponent,
+    ProcessorPauseGateComponent, ProcessorReadyBarrierComponent, ProcessorUniqueId,
+    ShutdownChannelComponent, StateComponent, SubprocessHandleComponent, ThreadHandleComponent,
 };
 use crate::core::processors::{PROCESSOR_REGISTRY, ProcessorInstanceFactory, ProcessorState};
 
@@ -287,7 +287,7 @@ fn spawn_dedicated_thread(
                 };
 
                 let state = match node.get::<StateComponent>() {
-                    Some(s) => s.0.clone(),
+                    Some(s) => s.clone_inner(),
                     None => {
                         tracing::error!("[{}] No StateComponent", proc_id_clone);
                         return;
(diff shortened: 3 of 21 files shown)


tato123:
## CodeRabbit review — addressed in ed279a88

**🟠 Major — Manual processors reporting Running before `start()` succeeds: valid, partially fixed.**

Verified and it was real. `start()` is Manual mode's actual bring-up, and Manual is what every Python processor's app-side host declares — so this hit the exact processors the readiness signal exists to watch. It could fail, log a `warn`, and return without touching state (`thread_runner.rs:471`), and `:110` then overwrote it with `Stopped`. A failed camera and a clean shutdown were indistinguishable in `health`.

Fixed: a failed `start()` marks the processor `Error`, so does the untrusted-tier refusal beside it, and the closing `Stopped` no longer buries either (`transition_to_unless_already_failed`, checked under the same lock as the write so a concurrent failure is not overwritten). Test: `stopping_a_processor_does_not_bury_why_it_stopped`.

**Not** fixed: moving the Manual `Running` transition to after `start()` succeeds. `Running` is still set for every mode at `spawn_processor_op.rs:384` before the mode function runs, so a readiness wait landing between `setup()` returning and `start()` failing still reports success. Making that per-mode changes when pause, resume, health and the control plane observe a processor as live — an execution-model decision that belongs to the owner and the plan, not to this ticket. Flagged for the owner rather than done silently.

**🟠 Major — synchronize readiness snapshotting with teardown: acknowledged, declining.**

The substantive version of this was real and is already fixed earlier in the branch: the binding used to hold a strong `Arc<Runner>` across the entire 30–60s wait. It now upgrades only to read the states and drops the reference before waiting, so the exposure is a graph read and a `Vec` build.

Closing the remainder needs one of:
- holding the lifecycle mutex across the snapshot — which works, but takes the graph lock with the GIL held and trades a microsecond-window spurious error for a real deadlock surface (an engine thread holding the graph lock while needing this interpreter's GIL);
- new synchronization between teardown and readiness — a design change, and `run()` deliberately releases the lifecycle guard *before* `drop_engine_without_holding_the_gil` precisely to avoid holding it across a GIL-released call.

Neither is warranted for a race that needs `shutdown()` to land inside a microsecond window on a path no caller exercises concurrently (`__enter__` waits; `__exit__` shuts down). Declining the suggested regression test too — it would pass by not hitting the window. Documented on `engine_to_read_graph_readiness_from` instead.

**🟡 Minor — document lifecycle `RuntimeError` cases:** fixed, the stub now names the shut-down case alongside failure and timeout.

**🟡 Minor — `raise ... from None` (Ruff B904):** fixed. Right call — the run-loop failure deliberately replaces the readiness error rather than chaining to it, since it is the cause and strictly more informative.

Verification: engine lib 1772 passed, wheel suite 174 passed / 1 skipped, harness suite green. One unrelated flake seen once in `stopping_an_already_stopped_runtime_is_a_no_op` — it subscribes to the process-global `PUBSUB`, which every `Runner::new()` re-initializes, so a non-serial test constructing a Runner concurrently wipes the subscription; `#[serial]` only guards against other serial tests. Pre-existing, passes 4/4 isolated and in two clean full runs.

Issue #1781, opened 2026-08-08
chore(docker): where the pre-pivot GPU container image lives in git history


tato123:
A pointer, so nobody has to spelunk. The Docker image story was deleted in #1712; this records where to read it and what is worth reading.

## Where

Last present on `main` at **`73078185`** (`docs(plan): propose the MCP-served-with-the-node change (#1780)`).

```
git show 73078185:Dockerfile
git show 73078185:docker-compose.yml
git show 73078185:docker/build-stage1.sh
git show 73078185:docker/entrypoint.sh
git show 73078185:docker/README.md

# or browse the whole thing:
git checkout 73078185 -- Dockerfile docker-compose.yml docker/
```

## What is worth reading, and what is dead

**Dead — do not try to revive it.** The bulk of `Dockerfile` (138 lines) and `docker/build-stage1.sh` is package-source machinery: it emitted an image-local `.slpkg` generic store + catalog + release manifest from the checkout, `streamlib link`ed the engine in via `[patch.crates-io]`, served the Deno SDK npm face over a `python3 -m http.server` mount, and let `runtime.add_module` compile the api-server from source on first boot. Every one of those mechanisms is deleted by the importable-Python-library pivot. Nothing in CI ever built the image.

**Worth reading — the image layer.** The CUDA runtime base pin, the apt package sets (GLVND/EGL dispatch, Vulkan loader, V4L2, userspace audio), and the entrypoint's PipeWire startup order. All of that is already captured, and now stated self-containedly rather than by reference, in [`docs/learnings/headless-nvidia-vulkan-container.md`](../blob/main/docs/learnings/headless-nvidia-vulkan-container.md) — read the learning first; go to the SHA only if you want the exact `RUN` lines.

## Still in the tree

Neither of these was deleted; both survive the pivot unchanged because they are about the host and the hardware, not packaging:

- `scripts/docker/host-prereqs.sh` — NVIDIA driver check, Container Toolkit wiring, `v4l2loopback` virtual camera nodes
- `docker/pipewire/10-virtual.conf` — declarative virtual null sink/source for a container with no `/dev/snd`

`docker/README.md` now describes exactly this state.

## Not a request to build anything

[...]

PR #1782, opened 2026-08-08
feat(runtime)!: relocate the api-server and trim the control plane to observation-only verbs

tato123:
## Summary

PR 1 of 2 on #1712. The api-server becomes engine-side infrastructure, and the control plane's vocabulary becomes observation-shaped.

**Relocation.** `packages/api-server` → `runtime/streamlib-api-server`. It is a host — statically linked, `crate-type = ["rlib"]`, never dlopen'd — not a distributable package. Both trees are two levels deep so the crate's own `../../` path deps resolve unchanged; the schema patch override becomes `../../packages/core`. Its `packages/*` facade-dep allowlist entry in `check-boundaries` retires with it, since that ratchet scans `packages/*` only.

**Observation-only trim.** A node's graph is defined by its code and the edit loop is `dev`, so the live-mutation verbs come out of both front ends and the CLI:

| surface | removed |
|---|---|
| REST | `POST /api/processor`, `POST /api/processor/source`, `POST /api/processor/source/replace`, `DELETE /api/processors/{id}`, `POST /api/connections`, `DELETE /api/connections/{id}` |
| MCP | `submit_processor`, `replace_processor`, `remove_processor`, `connect` |
| CLI | `submit`, `replace`, `connect`, and `remove`'s control-plane mode |

`src/ops.rs` — the shared submit/replace composite both front ends called — goes whole, with its wire types, error bodies, and the `ProcessorLanguage` DTO that served only them. **`shutdown` and `health` survive**: the plan names exactly four verbs for removal and neither is one. `remove` keeps its package mode, which belongs to the module system and is #1715's.

**Docker.** Deleted, not rewritten. The image built an image-local `.slpkg` tree, `streamlib link`ed the engine in via `[patch.crates-io]`, served the Deno SDK over an HTTP mount, and let `runtime.add_module` compile the api-server from source on first boot — every one of those mechanisms dies with this pivot, and nothing in CI built it. `scripts/docker/host-prereqs.sh` and `docker/pipewire/10-virtual.conf` stay: they are host/hardware, not packaging. The durable image-layer knowledge was already in `docs/learnings/headless-nvidia-vulkan-container.md`; its two "Applied in `<file>`" pointers would have dangled, so each is replaced by the fact itself and the learning now stands alone.

**`schemas.yml`.** Re-scoped to its OpenAPI half per the owner's 2026-08-05 call. The JTD → TypeScript → Zod half is what §Processor model deletes and what has failed since 2026-01-08. `workflow_call` goes too, since release-please already refused to call it.

## Closes

Partially addresses #1712 (PR 2 completes it)

## Exit criteria

- [x] api-server lives in `runtime/`, every path dep and boundary allowlist updated
- [x] No graph-mutation route or MCP tool is served, documented, or reachable from the CLI
- [x] `graph` / `tap` / `logs` / `shutdown` / `health` / `nodes` unchanged, auth gate intact
- [x] Docker packaging removed with the host layer preserved and a history tracker filed (#1781)
- [x] `schemas.yml` generates and publishes the OpenAPI client only

## Test plan

**Two-sided absence guards**, on both front ends. The MCP catalog is asserted **equal** to the observation vocabulary rather than a superset, so a tool that regrows fails even if nothing references it; a companion test drives every deleted tool name through `tools/call` and requires "unknown tool". On REST, every deleted route must be unrouted — checked with auth **off**, so a 401 can never be mistaken for absence — and absent from the OpenAPI spec. Both stubs `unreachable!` on every graph-mutating `RuntimeOperations` method: the trait is unchanged, but no route or tool may reach one.

`cargo test`: api-server 44, cli 87 + 18 integration, processor-extract 137, xtask 293, streamlib-runtime boot 4. Gates: `check-boundaries`, `check-manifest-schema`, `check-processor-source-reachability`, `check-package-version-drift`, `lint-logging`, `check-pack-load`, `install-packages` (this one exercises the relocated schema patch path end to end). Reviewed by `review-pr` and `rust-craftsmanship-reviewer` over two rounds; both APPROVE.

Four pre-existing failures on `main` are unchanged and unrelated: `--all-features` workspace check/clippy (blocked by `test-fixtures-abi-mismatch`'s mutually-exclusive `tamper-*` features and a stale nightly gate in vendored vulkanalia), plus clippy in `streamlib-idents` and `setup.rs`.

## Notes for owner

**A duplicate OpenAPI inventory, found in review.** `src/bin/generate_openapi.rs` carried a *second* hand-written spec with dummy handlers, and `schemas.yml` publishes exactly that one — so the generated TypeScript client would have kept advertising routes the node no longer serves. Rather than prune the four stale stubs I collapsed the duplication: route registration now lives in `always_open_routes()` / `bearer_gated_routes()`, which both `build_router` and a new `control_plane_openapi_spec()` compose, and a test asserts the generated and served documents are byte-equal. This also fixed drift predating the branch — the generated spec had been missing the shutdown route, both schema routes, and the tap WebSocket.

**A `std::thread::sleep` in `boot.rs`, flagged so you have the hook.** The `/ws/events` regression lock used to trigger its event by POSTing to `/api/processor`. That route is gone, and the only surviving route that publishes is `shutdown` — which ends the node ~50 ms later, so the trigger is one-shot where the old one was endlessly repeatable. `PUBSUB.subscribe` returning is not a readiness signal (it spawns a thread that opens its iceoryx2 service with up to 10 retries at 20 ms), so a readiness frame from the handler would not substitute. The test now gives that documented budget an order of magnitude of headroom, then triggers once. `engine-doctrine.md` says timekeeping is "never sleep-based"; I read that as governing how the runtime measures time rather than a test's readiness wait, but if you read it as absolute, this is the line. Verified 49 green runs including 16-way concurrency.

**A production change I considered and declined.** A readiness frame on `/ws/events` would be a real API improvement — today no consumer can know when its subscription goes live — but it does not fix the race above and this ticket is not chartered to change that wire contract. Its own ticket if you want it.

**Expected downstream breakage, not filed as tickets** per the packages-lag doctrine: `examples/api-server-demo` exercises all six deleted REST routes and will not build.

**PR 2 is blocked on nothing but this.** It retires `tools/streamlib-cli`, deletes the `mcp` verb and `serve_stdio_jsonrpc` per the merged mcp-served-with-the-node change, and ports `nodes` / `graph` / `tap` / `logs` into the wheel.

## Summary by CodeRabbit

* **New Features**
  * Added control-plane APIs for graph inspection, schema and registry discovery, event streaming, channel taps, logs, and runtime shutdown.
  * Added on-disk node discovery and OpenAPI specification generation and serving.
* **Changes**
  * Control commands and MCP tools now focus on observation and shutdown; graph mutation commands were removed.
  * Replaced the self-contained Docker image workflow with wheel-based deployment guidance.
* **Documentation**
  * Updated architecture, deployment, NVIDIA Vulkan, and host prerequisite documentation.


tato123:
Thanks — 4 accepted in `5abfc4ad`, 4 declined with reasons.

## Accepted

- **`host-prereqs.sh:89` (Major)** — you caught a regression this branch introduced. The old closing line named the next command; I rewrote it into an assertion of readiness while the GPU-passthrough check only warns and exits 0. `warn` now counts and the closing line reflects the count.
- **`schemas.yml:72-73`** — pinned `openapi-typescript@7.9.1`. This job publishes a package; an upstream release must not change generated output between two dispatches of the same commit.
- **`install_packages.rs:289`** — `skip` / `dist` → `skipped_path_dependency_package` / `distributable_package`.
- **`generate_openapi.rs:31`** — partially. Dropped the "Created directory" line (`create_dir_all` is already idempotent, so the `exists()` guard went too). Kept the one line naming where the artifact landed — see below.

## Declined

**`schemas.yml:90-95` (Major) — preserve the published package exports.** There is nothing to preserve. `schema-types` and `schema-validators` were generated from the JTD schema layer, and the plan deletes that layer outright (`ARCHITECTURE.md` §Processor model: *"no JTD, no schema registry, no embedded schemas, no codegen and no generated type classes"*). A deprecated compatibility entry point would have to export types no generator can produce. Compounding: the package last published successfully on 2026-01-08 and has failed every release since, nothing in the tree consumes it, and the workflow is now `workflow_dispatch`-only — so there is no release for a migration path to sit on. The repo is also pre-1.0 with an explicit no-back-compat-shims rule. Owner call, 2026-08-05.

**`handlers.rs:356` (Major) — bound the event-forwarding channel.** Correct and worth fixing, but pre-existing and untouched by this branch — `handle_websocket` is carried through the relocation byte-for-byte. Engine doctrine says findings outside the ticket get reported, not folded into the diff. Flagging for the owner rather than filing: a slow `/ws/events` reader does grow that channel without bound.

**`handlers.rs:175-184` — log the discarded graph-export error.** Same disposition: `get_graph` is pre-existing and not part of the verb trim. Agreed on the merit — a bare 500 with no server-side record is bad, and the file's other error paths do record the cause.

**`node_registry.rs:129-148` — explicit names for registry I/O state.** File carried unchanged by the relocation; it appears in the diff only because git rendered the move as delete+add. Not this ticket's to reshape.

## One correction

On `generate_openapi.rs`: *"Do not suppress `clippy::disallowed_macros`"* is not this repo's convention. `#![allow(clippy::disallowed_macros)]` is the sanctioned allowance that the engine's own `lint-logging` gate reads and honours — `xtask/src/lint_logging.rs:861-875` matches that attribute explicitly, alongside a `streamlib:lint-logging:allow-file` pragma for non-Rust files. This is a build-time codegen binary with no `tracing` subscriber to emit through, so the attribute is the right mechanism, not a suppression of the rule. I've made the rationale on it say that rather than leaving it terse.

PR #1785, opened 2026-08-08
feat(cli)!: the observation verbs ship in the wheel, and MCP loses its second transport

tato123:
## Summary

PR 2 of 2 on #1712, completing it. PR 1 merged as `e9c4f137`.

**`streamlib nodes` / `graph` / `tap` / `logs` now ship in the wheel**, so a `pip install streamlib` user can inspect a node and not just launch one. Stdlib only — `urllib` for the JSON-RPC client, `os.kill(pid, 0)` for liveness, `json` for the registry and the log. Nothing new enters the wheel's dependency closure so a user can look at their own pipeline.

Three modules behind the argparse wiring:

| module | job |
|---|---|
| `_node_registry` | reads `$XDG_RUNTIME_DIR/streamlib/nodes`, liveness-checks both signals, prunes only when **both** say dead |
| `_control_plane_client` | resolves the target (`--url` → `--node` → sole live node) and marshals one `tools/call`, distinguishing all four ways a call fails |
| `_runtime_log_reader` | renders the JSONL, tails it, and switches on rotation |

**The verbs leave the Rust CLI** — keeping them in both would be the parallel system engine doctrine forbids. **MCP loses its second transport**: `serve_stdio_jsonrpc` had exactly one non-test consumer, and MCP is served by the node's own control plane at `POST /mcp`, mounted with the node and sharing its lifecycle. That takes the `for_stdio_protocol` → `PrettyMirrorStream` → `pretty_mirror_stream` chain with it, per the merged `mcp-served-with-the-node` change — **all five of its REMOVED patterns now reach zero in the tree.**

The log directory is resolved by the engine through a new `runtime_log_directory()` binding rather than recomputed in Python: those rules are the engine's, and a second implementation would drift into reporting "no logs" for a runtime that logged fine.

## Closes

Closes #1712

## Exit criteria

- [x] `nodes` / `graph` / `tap` / `logs` served by the wheel's console script, stdlib only
- [x] The same verbs gone from the Rust CLI; `mcp` and `serve_stdio_jsonrpc` gone entirely
- [x] Record rendering byte-identical to the engine's own mirror, held there by a test on both sides
- [x] All five `mcp-served-with-the-node` REMOVED patterns at zero

## Test plan

**The record rendering is a cross-language contract**, so it is locked from both sides: `core/logging/tests.rs` calls `format_event_pretty` for real and asserts the same golden literal the Python test asserts, and each names the other. Changing either alone goes red.

62 new hermetic Python tests — a real loopback HTTP server, a real temp registry, real files. Verified against the deleted Rust on a live 98-record log file: byte-identical output, and identical under every filter. Driven end-to-end against a real `streamlib dev` node: `nodes` listed it, `graph` returned its topology through both sole-node resolution and `--url`, a clean SIGINT took the entry away, and a synthetic dead entry was pruned.

Gates: wheel pytest 183 passed / 55 deselected (237 with GPU tests on the rig), pyright 0 errors, stubtest clean, api-server 41, cli 49, engine logging 30, `cargo check --workspace --all-targets` clean, 20 xtask lints green. Reviewed by `review-pr` and `rust-craftsmanship-reviewer` over three rounds; both APPROVE.

## Notes for owner

**A gate hole, larger than this PR.** `_engine.pyi`'s `runtime_log_directory` said `-> str` while the function returns `pathlib.Path` — and **stubtest passed anyway**. It checks presence and signature arity, not native return types. That means the `.pyi` return annotations on *every* `#[pyfunction]` are unverified by any gate, not just this one. Worth its own look.

**`- REMOVED: Commands::Shutdown` is missing.** Deleting the CLI's `shutdown` verb is plan-consistent (§Control plane's verb list excludes it), but no change file carries a REMOVED bullet for it, so `/ship-change` cannot prove the removal. `/implement` cannot touch `docs/plan/**`, so this needs a line added to `mcp-served-with-the-node.md`. **Related ruling wanted:** the `shutdown` MCP *tool* is still served with no CLI client, which sits oddly against "agents and humans use the same verbs" — retire the tool too, or leave it agent-only?

**Two deliberate narrowings**, both flagged rather than silent:
- **`tools/streamlib-cli` the crate survives.** I announced retiring it; it still owns `add`/`link`/`pkg`/`install`, and the `install-packages` CI gate drives that binary. Those verbs and that gate retire together under #1715 — deleting the crate here would take a live gate with it. Both reviewers agree this is the right call.
- **`--source` accepts `rust|python`, not `deno`.** New surface, and the Deno SDK is deleted by the plan. `Source::Deno` survives in the engine enum, so a pre-existing record carrying `"source":"deno"` is simply no longer selectable.

**A note, not a ticket:** `registry_directory()` is reimplemented in Python while the sibling log-dir resolver got a native binding — the same reasoning arguably applies to both. The fallbacks differ only when `XDG_RUNTIME_DIR` is unset (already the degraded path), and a second binding is new public surface plus a stub entry; not worth riding in a PR this size.

**Diagnostic, for #1713:** the wheel and the Rust CLI resolve the log directory to different places — the wheel lands in `…/uv/python/cpython-3.12/bin/.streamlib/logs`, because `find_app_root()` walks up from `current_exe()` looking for `packages/` and a pip-installed interpreter has no such ancestor. Correct-per-host today, but it is the same root cause as the red CI wheel job (`/usr/bin/.streamlib/logs` → permission denied). `importable-python-library-ripout.md` already assigns that resolver's re-scope to #1713.

## Summary by CodeRabbit

- **New Features**
  - Added Python CLI commands for discovering nodes, viewing graphs, tapping streams, and reading or following runtime logs.
  - Added control-plane connectivity, authentication, error handling, log filtering, formatting, and log rotation support.

- **Changes**
  - MCP serving is now HTTP-only; stdio serving is no longer available.
  - Runtime logs are mirrored to standard output consistently.

- **Breaking Changes**
  - Removed the Rust CLI’s observation, logging, MCP, and node-management commands.


tato123:
All 7 taken in `6511606f` — 4 inline and 3 nitpicks, no declines.

## Minor

- **Malformed response shapes (`_control_plane_client.py:205`)** — right, and worth more than it looked. Every member is now shape-checked, not just presence-checked. I also went one step further than proposed: a result that succeeds but carries no readable text now raises too. Returning `""` printed a blank line and exited 0, which reads as *"the node has nothing"* rather than *"this response made no sense"*.
- **`_decode_line` schema validity (`_runtime_log_reader.py:260`)** — correct, and the same class of bug as one this PR already fixed in `--list`: a record that decodes as JSON but isn't shaped like a record took the renderer down, ending the whole read over one bad line.
- **`--list` ignoring flags (`cli.py:497`)** — good consistency catch. The control-target path already treated a silently-dropped flag as a wiring error; the two modes disagreed on the same class of mistake. They now agree.
- **`__all__` sorting (RUF022)** — done.

## Nitpicks

- **`--help` in a loop** — hoisted; it was starting three interpreters to read the same output.
- **Scheme allowlist** — done. Also worth noting the second vector you named: it isn't only `--url`, a corrupt `control_url` read out of a registry entry reaches the same call.
- **`ObservationVerbUsageError`** — done, and I took the rename too: `read_runtime_log_file` → `render_runtime_logs`, since it also serves the `--list` enumeration, which reads no single file.

## Verification

12 tests added — non-http schemes, five misshapen 200 bodies, four schema-invalid records, the `--list` rejection. Suite is 195 passed / 55 deselected; pyright 0, stubtest clean, `lint-logging` clean.

Also driven against a live node: `graph` still returns its topology, `--url file:///etc/passwd` is refused before any request, `logs --list --level warn` is refused by name, and plain `--list` still lists.

PR #1820, opened 2026-08-10
feat(ipc)!: 76-byte frame header — delete the wire schema tag and the per-read comparison

tato123:
## Summary

Shrinks the IPC frame header from **204 to 76 bytes** by deleting the 128-byte `SchemaIdentWire` tag it carried, and deletes the per-read comparison that tag existed to feed. New layout: `port_key 0..64`, `timestamp_ns 64..72`, `len 72..76`, `data 76..`.

The tag had no reader left worth serving. The only thing that consulted it was a comparison in `read_raw_bounded` whose entire output was a warn-once `tracing` line with no production consumer — it never blocked delivery, never returned an error, never reached the graph. After this, no read path examines a type at all: a mismatch surfaces as a decode failure at the consuming processor.

**This is one commit's worth of wire break by necessity.** There is no version field in the header to negotiate on, so every publisher and subscriber moves together — the host's write and read paths, the compiler op that wires them, and the wheel's helper path. A partial migration would be silent cross-process corruption rather than a compile error.

Deleted: `SchemaIdentWire` with its error type, four size constants, compile-time ABI lock and slice codecs; `core/schema_agreement.rs` whole; `PortConfig`'s expected-ident and mismatch-latch fields plus their two accessors; `schema_ident_wire_for_spec` and `schema_ident_json`. `set_channel_publisher`, `wire_rust_source`, `wire_rust_dest` and `wire_subprocess_source` drop their schema parameters; the host→subprocess wiring envelope loses its `"schema"` key and the wheel stops parsing one.

Net **−783 lines** across 18 files, adding no abstraction.

## Closes

Closes #1814

## Exit criteria

- [x] `FRAME_HEADER_SIZE` is 76 with the layout above; no header field carries a schema ident
- [x] No read path compares a type — `schema_agreement.rs`, the comparison block and the mismatch flag are gone
- [x] The wiring envelope carries no `"schema"` key and no parser looks for one
- [x] Both movers consistent: the host and the wheel's helper path

## Test plan

**Layout regression test** — `frame_header_fields_sit_at_their_documented_wire_offsets` asserts every field's byte offset in the *serialized slice*, not the Rust struct layout. The ticket asked for a `#[repr(C)]` layout test; `FrameHeader` is not `#[repr(C)]` — it is packed field-by-field by `write_to_slice` — so the slice offsets are the actual wire contract. Ticket body corrected with a strikethrough to match. Independently verified red under two mutants: a re-added 128-byte gap, and a same-size timestamp/len reorder that the round-trip tests do *not* catch.

**Which test is the revert lock.** The tag was inert on the read path before deletion, so the new Python conformance tests pass against the tagged wire too — they pin the contract forward. The byte-offset assertion is what goes red if the tag comes back. Recorded in the test file so a later reader doesn't mistake one for the other.

**Cast-not-contract conformance** — four new tests over a real wired iceoryx2 link (verified live, not a stand-in: matching channel names deliver, mismatched names return `None`): a producer publishing type X reaches a consumer reading type Y; the mismatch surfaces only at a read that asked for strictness; that read *consumes* its bag rather than leaving it queued to re-raise forever; and the link keeps delivering afterwards.

Run green: `streamlib-ipc-types` 13/13 · engine `--lib` iceoryx2/embedded_schemas/open_iceoryx2 84/84 · `streamlib-processor-schema` 28/28 · wheel `test_read_into_target` + `test_helper_process` 32/32 · full local gate battery **60 passed / 0 failed** (engine lib 1211, wheel pytest 211, stubtest, pyright).

## Notes for owner

**1. `docs/architecture/schema-identity-and-packaging.md` is now wrong — and correctly so; no action wanted.** Decision 2 still lists "iceoryx2 payloads" as a schema-carrying wire surface, which this PR falsifies. I initially flagged this as needing a supersession annotation; that was wrong. The doc is already commissioned for **deletion**, not annotation — owner ratification at the #1715 `/ship-change` gate (2026-08-09) kept it alive only "until its subject dies in the successor changes", naming `schema-free-ports` and `processor-class-identity`; `OPERATING-MODEL.md` §4 resolved the same on 2026-08-02. Its edit-deny rule is doing exactly its job: all seven of its architectural decisions describe machinery the plan deletes wholesale, so per-ticket annotation would be churn on a doc scheduled to go. **The thing to make sure actually happens is the deletion, at the `/ship-change` for whichever of the two chains lands last.**

**2. `cargo fmt --all` is not usable on this repo, and no CI gate runs it.** Main is not rustfmt-clean under stable rustfmt 1.8.0 — 88 files differ on a pristine tree — and no workflow runs `cargo fmt` or `clippy`. I ran it once early, it touched 74 files, and I reverted all collateral; this diff is semantic-only, no whitespace hunks. Flagging because it is a trap for the next session.

**3. Two pre-existing hot-path costs — now ticketed as #1821 and #1822, both blocked by #1816.** Neither is fixed here (deletion ticket; report-don't-auto-fix). One correction to what I first wrote: the unchecked-length panic is **not reachable from correct code** — `write_raw` always stamps `len` from the payload it just sized for, so every producer today is self-consistent. It is trust-boundary hardening, not a live crash, which is why it is filed as a ~1-line fix separate from the loan-path rework rather than bundled with it.
   - **#1821** — `read_raw_bounded` slices at the length the header claims without checking the frame carries it; `len` became the header's entire trust surface when this PR deleted the tag.
   - **#1822** — `write_raw` allocates a staging buffer and copies the frame into it before copying it *again* into the zero-copy loan; the read side pays two more copies. At 64 KiB payloads that is ~256 KiB per frame through allocator and memcpy, on a transport chosen for the loan.

**4. `docs/testing-hardware.md`'s exclude list is stale** — all five crates it names are no longer workspace members.

**5. What the ship gate sees now.** All six REMOVED patterns this ticket owns are green. Six remain red in `schema-free-ports.md` — `PortSchemaSpec`, `embedded_schemas`, `flow_class`, `streamlib-jtd-codegen`, `__streamlib_schema_ident__`, `check-processor-spec-new` — all belonging to later tickets in the same change. `sdk/streamlib-python-wheel/python/streamlib/schema_ident.py` also still exists, correctly out of this lane; it is the next thing a reader will ask about.

**6. One deliberate scope call.** Collapsing the wire resolver's dead `direction` parameter exposed that it was a verbatim copy of `embedded_schemas::resolve_node_port_schema`, whose own doc already claimed to be "the single graph→port-schema primitive" behind both call sites. That sentence was false; this PR makes it true (search first, extend never parallel). A degraded `PortSchemaSpec::Named` still fails the link at wire time — now as a typed `Configuration` error naming the unregistered schema rather than a panic. The panic that went away only guarded the *destination* spec, which the plan says must not be consulted at all.

## Summary by CodeRabbit

* **Changes**
  * Frame headers are now smaller and contain port, timestamp, and payload length only.
  * Output links and publishers no longer accept or transmit schema metadata.
  * Schema-mismatch validation has been removed from delivery paths.
  * Data delivery remains type-agnostic; strict reads detect incompatible types at consumption time.
  * Python output-link APIs no longer include an optional schema argument.

* **Tests**
  * Added coverage for producer/consumer type mismatches, failed casts, and continued message delivery.


tato123:
Both findings verified as real, and the second one caught something I missed — thank you. Dispositioning them to **#1821** rather than this PR, with reasoning below.

### Verified by execution, not review

I extracted `PortKey` + `FrameHeader` verbatim from this branch and ran the malformed inputs:

| probe | result |
|---|---|
| `read_from_slice(..).port()`, prefix `0xFF` | **panic** — `range end index 255 out of range for slice of length 63` |
| `read_port_from_slice`, prefix `0xFF`, 76-byte frame | **panic** — `range end index 256 out of range for slice of length 76` |
| `read_port_from_slice`, prefix `0xFF`, 300-byte frame | no panic — silently returns **255 bytes of payload as the port name** |

That third row isn't in your report and is arguably the worse one: on any frame big enough to absorb the over-long prefix there's no crash at all, just misrouting.

### Why not in this PR

Both sites are **pre-existing and unmodified by this PR**:

- `read_from_slice` — `origin/main` carries the identical unclamped `port_key.len = buf[0];`. This PR's only change there was rewriting it as a struct literal to clear a `field_reassign_with_default` clippy warning. Same semantics, same exposure.
- `read_port_from_slice` — not touched by this diff at all.

Neither is reachable from correct code today: `PortKey::new` rejects a name over 63 bytes and the writer sizes each frame to the length it stamps, so every producer is self-consistent. And `.port()` is never called on a wire-read header in production, while `read_port_from_slice`'s only caller is `route()` — a test/fixture injection path, since the live receive path routes by subscriber binding rather than by the header.

Your framing that *"the frame now carries no schema field, so the length prefix is the only variable-width field left on the read path"* is a fair argument for why this matters **more** after this PR — it's the same argument I made when filing #1821 — but it isn't evidence this PR introduced it. This is a deletion ticket, and repo doctrine is to report surfaced-but-unrelated issues rather than fold them into an unrelated diff.

### Where it went

Your line *"One validation policy at the deserialization boundary covers both sites"* is exactly right, and it's why I widened rather than split. **#1821** was filed before this review covering only the payload-length site; it now covers all three, retitled to `fix(engine): bound every wire-derived length the frame-header read path trusts`, and carries your panic evidence plus the silent-misrouting case.

One deliberate divergence from your proposed diff: clamping is right for the port-key prefix (a name can't exceed the field holding it, so a clamp is lossless for well-formed frames and yields a harmless non-matching port for malformed ones), but wrong for the payload length — a frame claiming more than it carries is unusable, and silently truncating it would hide the corruption. That one should fail as a typed error naming the port and both numbers. #1821 specifies both shapes.

PR #1826, opened 2026-08-11
fix(build): realign intra-workspace version pins to 0.17.0

tato123:
## Summary

`main` does not compile. The 0.17.0 release commit (`63666170`) bumped `[workspace.package] version` but left every intra-workspace path-dependency pin at `version = "0.16.0"`, so Cargo cannot resolve the workspace and no crate builds.

This realigns those pins. Mechanical only — no dependency added, dropped, or re-pointed.

Scoped to the **17 workspace-member manifests** that carry a stale pin (49 pins total). The vendored vulkanalia crates are workspace members but carry no pins and are untouched.

## What was broken

| Probe | Before | After |
| --- | --- | --- |
| `cargo metadata --format-version 1` | fails — `failed to select a version for the requirement streamlib-processor-schema = "^0.16.0"` | resolves |
| `cargo check --workspace --all-targets` | cannot start | clean |
| `cargo xtask check-boundaries` | red | pass |
| `cargo xtask check-schema-versions` | red | pass |
| `cargo xtask check-device-wait-idle` | red | pass |
| `cargo xtask check-processor-spec-new` | red | pass |
| `cargo xtask check-no-escalate-in-lifecycle` | red | pass |
| `cargo xtask check-no-streamlib-metadata` | red | pass |

The commit immediately before the release (`f27cf414`) resolves fine, which isolates the break to the release commit. The only two green checks on `63666170` — License Header Check and Ship-Change Removed Gate — are the two that never invoke cargo.

## Closes

None — this is the mechanical unblock, not the fix.

Refs #1825.

## Exit criteria

- [x] `cargo metadata --format-version 1` resolves
- [x] `cargo check --workspace --all-targets` finishes clean (warnings pre-existing, 0 errors)
- [x] All six previously-red cargo-based xtask gates pass
- [x] `vendor/tatolab-vulkanalia*` untouched
- [x] No dependency added, removed, or re-pointed — version strings only

## Test plan

Ran locally on the branch: `cargo metadata`, `cargo check --workspace --all-targets`, and each of the six xtask checks individually. Results in the table above. Nothing here is behavioural, so there is no new test — the gates that could not run are the test.

## Notes for owner

- **This is not the fix.** The root cause is release configuration: `release-please-config.json` declares a single package (`"."`) and sets no `plugins`, so nothing is aware of the 49 pins. 0.18.0 breaks identically. Filed as #1825 with the two candidate fixes (add the `cargo-workspace` plugin, or drop the redundant `version` key from path deps) — that is a distribution decision against `ARCHITECTURE.md` §Distribution & versioning, so I did not pick one.
- **Diagnostic trap worth knowing:** `cargo metadata --no-deps` reports *success* on the broken tree because it skips resolution. The failure also names an unrelated crate pair, so it reads as "the branch I just cut is broken" rather than "main is broken". Cost me a detour; recorded in #1825.
- `packages/test-fixtures` is edited here. It is a workspace member and is one of the three `packages/` entries CLAUDE.md exempts from the consumer deny-list, so this is engine-tree work, not a consumer edit.
- `Cargo.lock` moves with the pins, as expected.
- Found while starting #1815 (delete the embedded schema registry); that work is branched and waiting on this to land.

## Summary by CodeRabbit

- **Chores**
  - Updated Streamlib components and adapters to version 0.17.0.
  - Updated SDK, runtime, media, graphics, CUDA, Vulkan, OpenGL, Skia, and test-support integrations.
  - Aligned Linux-specific and optional integrations with the 0.17.0 release.
  - No user-facing API or behavior changes were introduced.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4882,7 +4882,7 @@ checksum = "6ce2be8dc25455e1f91df71bfa12ad37d7af1092ae736f3a6cd0e37bc7810596"
 
 [[package]]
 name = "streamlib"
-version = "0.16.0"
+version = "0.17.0"
 dependencies = [
  "iceoryx2",
  "serde",
@@ -4892,7 +4892,7 @@ dependencies = [
 
 [[package]]
 name = "streamlib-adapter-cpu-readback"
-version = "0.16.0"
+version = "0.17.0"
 dependencies = [
  "libc",
  "serde",
@@ -4911,7 +4911,7 @@ dependencies = [
 
 [[package]]
 name = "streamlib-adapter-cuda"
-version = "0.16.0"
+version = "0.17.0"
 dependencies = [
  "cudarc",
  "dlpark",
@@ -4931,7 +4931,7 @@ dependencies = [
 
 [[package]]
 name = "streamlib-adapter-opengl"
-version = "0.16.0"
+version = "0.17.0"
 dependencies = [
  "gl",
  "khronos-egl",
@@ -4954,7 +4954,7 @@ dependencies = [
 
 [[package]]
 name = "streamlib-adapter-skia"
-version = "0.16.0"
+version = "0.17.0"
 dependencies = [
  "libc",
  "libloading 0.8.9",
@@ -4977,7 +4977,7 @@ dependencies = [
 
 [[package]]
 name = "streamlib-adapter-vulkan"
-version = "0.16.0"
+version = "0.17.0"
 dependencies = [
  "libc",
  "serde",
@@ -5022,7 +5022,7 @@ dependencies = [
 
 [[package]]
 name = "streamlib-consumer-rhi"
-version = "0.16.0"
+version = "0.17.0"
 dependencies = [
  "libc",
  "streamlib-surface-client",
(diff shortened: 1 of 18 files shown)


tato123:
CI status: every check green except **Wheel build + interpreter lifecycle (no GPU)**, which is pre-existing and unrelated.

The failure is `core::runtime::runtime::tests::stopping_an_already_stopped_runtime_is_a_no_op` — *"RuntimeStopping must be published exactly once across two stops"*. Nothing to do with version pins.

Verified by name rather than assumed: the identical test fails on `main` at `f27cf414` (run [31446669346](https://github.com/tatolab/streamlib/actions/runs/31446669346)), the commit immediately **before** the release — i.e. the last main commit that could still compile and reach this step. `main` at the release SHA cannot reach it at all, which is what this PR fixes.

So: 17 of 18 checks green, and the 18th is a known-red engine test that this branch neither introduces nor touches.

tato123:
Stacked follow-up now open: #1827 (`feat(engine)!: delete the embedded schema registry`, #1815) is based on this branch and will retarget to `main` when this merges.

This one should go in first regardless of #1827 — `main` does not currently compile for anyone.

Issue #719, opened 2026-05-07
feat(examples): generate structured SchemaIdent literals from sibling streamlib.yaml at build time


tato123:
## Description

Polyglot Rust example crates that spawn Python/Deno subprocess processors hand-roll a structured `SchemaIdent::new(Org::new(...), Package::new(...), TypeName::new(...), SemVer::new(...))` literal in `main.rs` whose segments mirror the sibling `python/streamlib.yaml` (or `deno/streamlib.yaml`) `package: { org, name, version }` block. When the YAML's `package.version` is bumped, the hardcoded `SemVer::new(...)` literal in `main.rs` doesn't move — the runtime composes its registered ident from the YAML, the host-side `add_processor` constructs a different ident from the literal, and the `add_v` lookup fails at runtime with no compile-time check to catch it.

Surface a build-time mechanism (build.rs reading the sibling YAML, or a const macro that reads it via `include_str!`) so the example call sites pull the same fields the runtime registration uses, and bumping the YAML version triggers a recompile rather than a silent runtime miss.

## Context

Surfaced in the #707 PR review as a drift risk on the polyglot example sweep (#712). The Rust-host-spawning-foreign-subprocess pattern can't reuse the `#[streamlib::processor]` macro's `<Module>::schema_ident()` form (the processor's Rust module doesn't exist in-tree), so a different mechanism is needed.

To the best of our current knowledge as of pickup the affected examples are:

- `examples/polyglot-continuous-processor/src/main.rs`
- `examples/polyglot-cpu-readback-blur/src/main.rs`
- `examples/polyglot-cuda-inference/src/main.rs`
- `examples/polyglot-dma-buf-consumer/src/main.rs`
- `examples/polyglot-manual-source/src/main.rs`
- `examples/polyglot-opengl-fragment-shader/src/main.rs`
- `examples/polyglot-vulkan-compute/src/main.rs`
- `examples/polyglot-vulkan-graphics/src/main.rs`
- `examples/polyglot-vulkan-ray-tracing/src/main.rs`
- `examples/camera-deno-subprocess/src/main.rs`
- `examples/polyglot-skia-canvas/src/main.rs`
- `examples/camera-rust-plugin/src/main.rs` (Rust dylib plugin variant)
- `examples/camera-python-subprocess/src/main.rs`
- `examples/camera-python-display/src/linux.rs`

Verify against current code at pickup — these are the files that landed structured-ident literals in #707 / #711 / #712.

## Architectural decisions locked

[...]

PR #1831, opened 2026-08-11
feat(sdk)!: strip the type from port declarations and control-plane rendering

tato123:
## Summary

The last step of `schema-free-ports`: a port declares name, description, and — on an input — delivery profile, and nothing else. Type information belongs to the authoring language and never reaches the engine.

Deleted end to end:

- **`PortSchemaSpec`** and its four impls, the `port_schema_spec_wire` msgpack mirror, `PortDescriptor.schema`, `ProcessorPortSchema.schema`, and `SchemaIdentOutput::from_port_spec`. `SchemaIdentOutput` itself survives — the MoQ catalog still uses it.
- **The positional `<schema>`** in the Rust `#[processor]` port grammar, plus `parse_port_schema` and codegen's `port_schema_spec_tokens`.
- **The `schema=` kwarg** on the wheel's `@input`/`@output`, `_resolve_schema_ident`, `streamlib/schema_ident.py` and its re-export, and the native `port_schema_spec_from_declaration`. Deleting `_resolve_schema_ident` is what retires `__streamlib_schema_ident__` — #1813 removed everything that emitted the marker, this was its last consumer.
- **`PortInfo.data_type`**, `PortInfoOutput.data_type`, `PortDescriptorOutput.schema`. A port renders as `{name, description, port_kind, delivery_profile}`; `PortInfo` gained the `description` it never carried and both DTOs gained the delivery profile the plan says port rendering includes.
- **`check-processor-spec-new`** — the 595-line xtask, its `Commands` variant and dispatch, and its workflow.

Net −2250/+330 across 42 files.

### Consequences forced by the deletion

These are not in the ticket's bullets; the tree does not compile (or would keep a no-op) without them:

- `ProcessorInstanceFactory`'s `RwLock<HashSet<PortSchemaSpec>>` — a surviving schema registry — with `known_schemas` / `is_schema_known` and its rebuild-on-removal pass.
- `ProjectConfig::resolve_bare_schema_refs` and `resolve_named_to_ident`. With no `Named` port spec left to rewrite, `load_with_link`'s `link_checkout` argument steers nothing, so it collapses back into `load`.

### One consequence that was *not* forced

`/api/registry` loses its `schemas` key, along with the `SchemaDescriptorOutput` / `SchemaFieldOutput` / `LinkBufferReadModeOutput` DTO chain. Only the list's *contents* were forced — its sole producer was `known_schemas()`. The three DTOs would have compiled untouched. They are removed because the list can now only ever be empty, not because it wouldn't build. Flagged below.

## Closes

Closes #1816

## Exit criteria

- [x] No port declaration in either authoring language accepts a type; `PortSchemaSpec` is gone from the tree.
- [x] `graph`, `tap` and every snapshot render a port as `{name, description, port_kind, delivery_profile}` — no type field, not even omitted-when-empty.
- [x] `check-processor-spec-new` and its workflow are gone.
- [x] The `streamlib-processor-schema` survivors have a home and the workspace builds.

**Crate layout.** The change file delegates layout to implementation with one constraint — *extend an existing crate, never stand up a new abstraction beside it*. `sdk/streamlib-processor-schema` stays and is stripped of its data-type layer only. It is the sole crate both `streamlib-macros` (a proc-macro, so it cannot depend on the engine) and `streamlib-engine` can share, and the survivors already live there. No new crate, module, or trait is introduced.

## Test plan

| Gate | Result |
|---|---|
| `cargo check --workspace --all-targets` | clean |
| `cargo test -p streamlib-engine --lib` | 1177 passed, 0 failed |
| `cargo test -p streamlib --lib` (incl. the example E2E) | 2 passed |
| `cargo test -p streamlib-processor-schema -p streamlib-macros -p xtask` | 262 passed |
| engine integration tests touched | 24 passed |
| wheel `pytest -m "not requires_gpu"` | 202 passed |
| `pyright` | 0 errors |
| xtask lint gates | all pass |
| `rustfmt --check` over touched files | clean |

The wheel's native module was rebuilt with `maturin develop` before the pytest run, so the Python→native `read_port_descriptors` path is exercised for real rather than against a stale `.so`.

New locks, in place of the tests whose subject was the type layer:

- `port_info_output_renders_exactly_the_declared_keys` asserts the exact key set; `port_info_output_carries_no_type_key_under_any_spelling` and the descriptor sibling assert absence under four spellings, so a type field cannot return under a new name.
- `a_positional_port_schema_is_an_error` / `a_bare_any_port_schema_is_an_error` — the old Rust form fails the parse rather than being tolerated.
- `test_processor_port_rejects_a_schema_key` — `deny_unknown_fields` makes a manifest port type a loud parse failure.
- `test_a_port_declaration_takes_no_schema` — `schema=` raises `TypeError: unexpected keyword argument`.

## Notes for owner

**1. `/api/registry` response shape changed, and my first justification for it was wrong.** As above: the three DTOs weren't forced out by the compiler, they're removed as permanently-dead surface. The ticket's Done-means covers `graph`, `tap` and snapshots — not `/api/registry`. `dist/schemas/openapi.json` is untracked and regenerated, but the generated `@tatolab/streamlib-client` TS types change shape. Say the word if you'd rather this rode Change 2.

**2. The ticket's ship-gate sequencing note is wrong, in our favour.** It says the `REMOVED` bullets can't go green until #1672 moves `packages/` and `examples/` out. The gate already excludes both trees outright, so they were never the blocker. All three bullets — `PortSchemaSpec`, `check-processor-spec-new`, `__streamlib_schema_ident__` — return **zero** hits under the gate's own exclude set on this branch. The real blocker was `docs/architecture/schema-identity-and-packaging.md`, fixed here.

**3. `check-processor-spec-new` was retired on a wrong premise.** The ticket says it "polices `SchemaIdent` literals in port specs." Neither of its two passes did. Pass 1 matched `ProcessorSpec::new("PascalCase"` (processor identity, #707) and is now redundant by types — `ProcessorSpec::new` takes `impl Into<ProcessorTypeReference>` whose only impl is `From<SchemaIdent>`, so a bare string no longer compiles. Pass 2 matched hand-rolled `SchemaIdent::new(Org::new(...))` literals under `examples/*/src/` (#719) and has **no** type equivalent — that rule is now unguarded. Its target tree is deny-ruled and lagging so the practical loss is small, but it's a live rule silently dropped. Worth correcting the bullet's justification at `/ship-change`.

**4. A port type survives in the MoQ catalog — deliberately untouched.** `MoqCatalogTrackEntry.schema: Option<SchemaIdentOutput>` is documented as "the schema identifier for the data on this track" and `catalog_entry_for_output_port` takes an output port's schema. It's inert (the one production caller passes `None`). I left it: the ticket doesn't name it, and networking is **OPEN** in the plan, so deciding whether "no port carries a type in any snapshot" reaches a MoQ track catalog would be making architecture. Your call whether it dies here or in a later change.

**5. `examples/hello-streamlib/src/hello_forward.rs` is edited, with your explicit grant.** `sdk/streamlib-sdk` `#[path]`-includes it into its own lib-test build, so it compiles inside an engine crate and would not build under the new grammar — the same carve-out and reason as #1817. The edit is exactly the two positional schema args, nothing else. `examples/**` stays deny-ruled otherwise.

**6. `PortInfo` has largely lost its reason to exist** (surfaced by review, out of scope). With `data_type` gone it is a strict projection of `PortDescriptor` plus `port_kind` — and `port_kind` is `Default::default()` at every construction site in the tree. Meanwhile `ProcessorInstanceFactory` keeps `port_info` and `descriptors` maps keyed identically and written in the same function. Engine doctrine's one-system-per-concern points at deriving `port_info` from `descriptors`, but that's a follow-on this ticket didn't ask for.

**7. Two `docs/plan/changes/schema-free-ports.md` line anchors were stale** and I corrected the ticket body with strikethroughs: `RegisteredPortReceipt.schema` no longer exists (it went out earlier in the stack), and the `processor_schema.rs:524, *not* :42` correction was inverted — the enum *is* at `:42`.

## Summary by CodeRabbit

* **New Features**
  * Simplified processor port declarations by removing explicit input/output schema requirements.
  * Registry and port metadata now focus on names, descriptions, delivery profiles, and port behavior.
  * Built-in media processors use clearer concrete video type handling where applicable.

* **Breaking Changes**
  * Removed schema-related port configuration and the `SchemaIdent` API from Python integrations.
  * Removed schema details from registry responses and processor descriptors.
  * Legacy schema arguments are now rejected in processor declarations.

* **Chores**
  * Removed obsolete schema validation tooling and CI checks.


tato123:
Went through both bots. **Three fixed in `963e634e`, four acknowledged and not fixed.**

## Fixed

**CodeRabbit — `test_input_port_retired_knob_rejected` passes for the wrong reason.** Correct, and the best catch of the batch. The fixture declared both `read_mode` and `schema:`; once a port `schema:` key became an unknown field, the parse failed on whichever `deny_unknown_fields` hit first, so the test would have stayed green if `read_mode` were quietly re-accepted. This one was mine by omission — I stripped `schema:` from the other fixtures in that file and missed this one. Fixture now carries only the retired knob, and the assertion names it rather than just checking `is_err()`.

**CodeRabbit — surviving `@input(schema=GeneratedClass)` claim (`schema-identity-and-packaging.md`).** Correct. I annotated the two bullets above it and left the tail of a third that still offered `schema=` as "an optional typed view". Superseded; the opt-in typed read is `ctx.inputs.read(port, into=T)`.

**CodeRabbit — Reference and Tests sections name deleted components.** Correct, and worse than reported: all four paths are gone, not just some. `sdk/streamlib-jtd-codegen/` (deleted by #1813), `runtime/streamlib-engine/src/core/embedded_schemas/` (#1815), and both `sdk/streamlib-python/python/streamlib/` entries. The Python test entry now points at `sdk/streamlib-python-wheel/tests/test_processor_declaration.py`, where those locks actually live.

## Acknowledged, not fixed

**GitHub code quality ×2 — "Keyword argument 'schema' is not a supported parameter name" (`test_processor_declaration.py:91,93`).** True, and the point. Both calls sit inside `pytest.raises(TypeError, match="unexpected keyword argument 'schema'")` — the test exists to prove `schema=` is gone rather than tolerated. Already carry `# type: ignore[call-arg]`.

**GitHub code quality — "Statement has no effect" (`:103`).** That's the `...` body of a port marker method. It's the established stub idiom throughout this file and the pre-existing tests; the bot flagged one of several identical instances.

**CodeRabbit — port schemas on `data_in` / `data_out`.** The anchor lands on a `use` statement in `processor_schema.rs`, but the analysis chain shows the real target is `packages/moq/streamlib.yaml:36,53`. `packages/` is deny-ruled and lags by design — an engine change breaking a consumer is expected, not a defect, and that tree is never edited to make an engine change pass. Out of scope here.

## Note

The wider body of `schema-identity-and-packaging.md` still describes the codegen pipeline at length (§ around lines 338-362, plus mentions at 10, 29, 93, 177, 507). That rot predates this PR — it came in with #1813 and #1815 — and rewriting those sections is a substantial doc change this ticket doesn't cover. Much of it dies wholesale in Change 2 (`processor-class-identity`), which removes `SchemaIdent`'s identity role. Flagging rather than expanding scope.

Re-verified after the fixes: workspace builds clean, `streamlib-processor-schema` 20 passed, and the ship gate's own grep still returns zero hits for all three `REMOVED` bullets.

PR #1850, opened 2026-08-12
refactor(engine)!: the registry, graph and control plane key on the import path

tato123:
## Summary

The **migrate** step of `processor-class-identity`. Everything that keyed on a processor's identity moves from the structured `SchemaIdent` to the import path of the class it is — the registry, the factory, the graph node, the spec, the snapshot, the pubsub events, the moq catalog, and the control plane's `type` field, which becomes a plain string instead of a three-key object.

After this, `GET /api/graph` renders `"type": "my_app.filters:BlurProcessor"`, and version-blind lookup has nothing left to be blind about.

**`ProcessorClassImportPath` is a validated newtype, not a `String`.** Per the owner's ruling folded into this ticket, the invariant has to survive assignment and deserialization, not just construction — so the inner string is private, `new` is the only constructor, `Deserialize` is hand-written to validate rather than inherit `String`'s, and there is no `Default`. Every inhabitant is valid, which is why the descriptor's field can stay `pub`.

**Validity is emptiness and nothing else.** The path holds two grammars at once — `my_app.filters:Blur` from Python, `my_app::filters::Blur` from Rust, discriminated only by the sibling `runtime` field — so there is no single grammar to check it against. The engine stores it verbatim and never parses it, which is what the plan requires.

**The swap hazard is closed.** `ProcessorDescriptor::new` took two adjacent `impl Into<String>` params — the identity and the description — so reversing them compiled and registered a processor under its own description. Reversing them is now a type error.

## Closes

Closes #1840

## Exit criteria

- [x] Registry, factory, graph, snapshot, pubsub events, moq catalog and control plane all key on the import path; no identity lookup consults org, package or version.
- [x] `GET /api/graph` renders `type` as a string — verified against the rendered payload and against the generated JSON Schema (`{"type": "string"}`).
- [x] `ProcessorTypeReference` and `SchemaIdentOutput` are gone from the tree (the ship gate now reports both, plus `schema_ident_any_version`, as removed).
- [x] An `rt.add` of an unregistered type still produces a visible node in `Error` state carrying the requested string verbatim.
- [x] `IsolationTier` no longer takes an `Org`.

## Test plan

| Gate | Result |
|---|---|
| `cargo test --workspace` (excl. wheel) | **1887 passed, 0 failed**, 152 ignored |
| `cargo test -p streamlib-python-wheel --lib` | 43 passed |
| Wheel pytest, `-m "not requires_gpu"` | **207 passed**, 61 deselected |
| `mypy.stubtest streamlib._engine` | no issues — the hand-written stub needed nothing |
| `cargo clippy --workspace --all-targets` | zero findings in changed files |
| `cargo doc -p streamlib --no-deps` | zero broken intra-doc links |
| rustfmt on changed files | clean |
| `xtask check-no-in-process-placement` / `-reverse-dns` / `-inventory-submit` | pass |
| `cargo check --target aarch64-apple-darwin` | engine-free crates pass; engine **not verifiable here** (see note 10) |

New locks:

- **Control-plane rendering** (`control_plane_processor_type_test.rs`) — asserts the rendered payload, not the DTO's Rust type, and its negative half walks every key at every depth so no `org` / `package` / `version` can come back beside the string.
- **Registry** (`processor_instance_factory.rs`) — registration and lookup by path, distinct paths do not collide, a duplicate is refused without overwriting, both registration paths share one key space, and two Rust processors sharing a module path are caught.
- **The newtype** — both grammars ride through verbatim, blank is refused, `Deserialize` refuses what the constructor refuses (over JSON *and* msgpack), equality is exact.
- **Python** — two classes from two modules in one graph register under two distinct paths.

## Notes for owner

Non-blocking findings, batched. Nothing here needs an answer before merge.

**1. `ProcessorDescriptor.name` is not fully vestigial, and #1841 needs to know.** The ticket's end state says it "becomes vestigial here". It very nearly is — nothing keys on it — but one reader survives: the plan DECIDES that a display name defaults to "the class's short name", this ticket forbids splitting the import path to recover one, and `name.r#type` is the only carrier of it. So `default_display_name` reads it.

I did **not** add a short-name field to the descriptor to close this, because that is new surface the change file does not declare and architecture is decided in the plan. Instead the handoff is now explicit: **[#1841 has a comment](https://github.com/tatolab/streamlib/issues/1841#issuecomment-5262257983)** recording that deleting the field means re-homing the short name, with both routes and the constructor-shape trap spelled out. The read is collapsed into one method (`ProcessorInstanceFactory::default_display_name`) so that ticket has a single seam to repoint. Tell me if you would rather the field land here instead — it is ~30 lines.

**2. A real defect fixed at the seam.** `register_dynamic` checked the `registrations` map while `register_descriptor_only` writes only `descriptors` and `port_info` — so a constructor-bearing registration silently overwrote a descriptor-only one under the same key, replacing its ports and descriptor with no error. Pre-existing, but it mattered little under a structured key and matters now that a Python class and a native built-in contend for one key space. Both paths now check `descriptors`. Found by one of this PR's own new tests.

**3. `/api/registry` has a second wire break, beyond the `/api/graph` one the ticket names.** A genuine planning gap: the plan named `ProcessorNodeOutput.processor_type` but not `ProcessorDescriptorOutput.name`, and deleting `SchemaIdentOutput` left that DTO with no landing type. It now carries the import path — the only answer consistent with `/api/graph` — and the field is **renamed `name` → `processor_class_import_path`**, so the same value is spelled the same way on both routes and the bare `name` stops failing the zero-context test. `get_registry` is the one route with a `body =` on the OpenAPI doc, so this is what a generated client sees.

**4. `ProcessorDescriptorOutput.entrypoint` now duplicates `name`** for every Python processor. `descriptor.entrypoint` is left alone here on purpose — #1839 forbids unifying identity with it — and it dies in #1841.

**5. The isolation-tier collapse went wider than "loses its `Org`", and I recorded it.** `is_cdylib_resident` has returned a literal `false` since the plugin ABI died, so with the org gone the derivation has exactly one reachable answer. Collapsing it takes the `STREAMLIB_SESSION_ISOLATION_TIER` operator knob with it — a knob that provably does nothing is worse than no knob. `IsolationTier`, its `Untrusted` variant and the `FullAccessGrant` moat all survive: the moat is a compile-time guarantee about who may mint an in-process `RuntimeContextFullAccess`, and `run_processor_loop` keeps its `isolation_tier` parameter and its untrusted arms compiled and tested. Added as `REMOVED:` bullets to the change file, per §Recording facts vs deciding.

**6. One test's meaning changed, not just its types.** `conflicting_destination_profile_is_a_configuration_error` registered two sinks distinguished only by *package* but handed both the same import path — after the re-key the second registration is a duplicate, so both destinations would have agreed on one profile and there would be no conflict to detect. They now carry distinct paths. Worth knowing that hand-written descriptor fixtures can collide this way; macro-declared processors cannot.

**7. The moq catalog's `schema` field is gone.** Nothing populated it — its only caller passes `None` — and a link declares no type for a catalog to publish.

**8. Residue for #1841.** `SESSION_ORG`, `Org::is_reserved_for_session`, `SchemaIdent::schema_identity_tuple` and `matches_schema_tuple` now have no non-test callers. Left in `streamlib-idents` rather than deleted piecemeal from a crate that ticket removes whole.

**9. Consumer breakage — expected, wanted, not fixed.** Rust consumers break on `processor_type_ref!` (20 examples) and `ProcessorSpec::new` (15 examples + `packages/camera/processors`). No Python consumer breaks: the `@processor` authoring grammar survives to #1841.

**10. The macOS cross-compile could not be completed in this environment.** `streamlib-processor-schema`, `streamlib-macros` and `streamlib-error` — where the new type lives — check clean against `aarch64-apple-darwin`. The engine and the wheel cannot: `iceoryx2-pal-posix`'s build script fails bindgen on a missing `libproc.h`, i.e. no Apple SDK on this machine. That is a dependency's build script and nothing in this diff reaches it, but the engine-tree half of the repo's cross-verification rule is genuinely unrun rather than passing. The place it would bite is `python_native_builtin_blocks.rs`, whose `CameraSource` / `DisplayWindow` arms are `#[cfg(target_os = "linux")]`; the return type was re-typed once for every arm rather than per-arm, so the non-Linux arms are the untouched `PyRuntimeError` paths.

**11. The Python identity test observes a log record, not the control plane.** The ticket's validation shape asks for a two-processor graph reporting both paths *through the control plane*, plus the helper reaching its class. What is covered device-free: the engine's own registration record, read back for two classes from two modules. The control-plane rendering itself is locked Rust-side against the exact `to_json_async` payload a client receives (`control_plane_processor_type_test.rs`), and the helper-reach half is rig-only — `test_helper_placement.py` is entirely `requires_gpu`, so nothing in CI can prove it either before or after this change.

**12. Two review findings I declined, with reasons.** Swapping the wheel's `entrypoint` read for the typed identity field would be *unifying identity with `entrypoint`*, which #1839 forbids and this ticket explicitly defers to #1841. And a typed `DuplicateProcessorClassImportPath` error variant would stop three tests grepping message prose, but the prose is the contract there — it names both languages' causes and their fixes, and asserting the variant alone would pass with an unhelpful message.

**13. Two notes the confirmation review surfaced, neither blocking.**

- `save_graph_snapshot` now asks the registry for the class short name to decide whether a node's display name is an override worth serializing, where it previously read the node's own ident. A node added *before* its type registered and saved *after* would emit a spurious `display_name`. Unreachable in practice — an unresolved node's snapshot cannot be loaded at all, and register-after-add is not a supported flow — but it is a behaviour difference from `main` and worth having on the record.
- The `/api/registry` and `/api/graph` wire *keys* differ (`processor_class_import_path` vs `type`) even though the value and type are now identical; the graph node renames its field because there it genuinely is the node's type. The Rust field names match, which is what stops the bare-`name` naming violation.

## Summary by CodeRabbit

* **New Features**
  * Processor identity is now represented by validated class import paths across graphs, snapshots, events, APIs, and registration.
  * Graph and catalog output serialize processor types as plain strings.
  * Python registration distinguishes classes by module and qualified name.
  * Processor display names use registered metadata with import-path fallbacks.

* **Breaking Changes**
  * Removed legacy processor-reference and version-agnostic identity macros.
  * Updated processor and application APIs to accept class import paths.
  * Removed provenance-based isolation configuration; engine-spawned processors use trusted installed isolation.


tato123:
## Review triage — 5 fixed, 3 declined

Pushed as `bf9ffae2`.

### Fixed

**The registration race (`processor_instance_factory.rs`) — the one that mattered.** Correct, and worse than the comment implies: both paths read `descriptors` to test for the path, dropped the read guard, then took fresh write guards, so two threads could both pass the check and the second would overwrite a live descriptor, port info *and* constructor. That defeats the first-registration-wins rule this PR introduced the duplicate check to hold. The `descriptors` guard now spans check-and-claim in both paths, and the struct documents it as the outer lock since a registration takes `port_info` and `registrations` inside it.

Worth noting the hazard was already known in the tree: `python_processor_registration.rs` holds a mutex of its own across check-and-register *because* of "the engine registry's non-atomic read-then-write". Nothing else did.

The new test races two threads through a barrier over 200 rounds. I verified it fails against the previous code rather than assuming — it reports `left: 2, right: 1`, both threads registered.

**Naming (4 sites), all real:**
- `producer_class_path` built `…::ConnectCheckProducer` while `PRODUCER_TYPE` was `"ConnectSilenceProducer"` — one concept under two names, introduced when this PR converted that fixture. Now derived from the constant.
- `structured_type` returns a quoted import path and nothing structured — renamed.
- Three local `import_path` helpers → `class_import_path`.

### Declined, with reasons

**Bump the MoQ catalog version and handle v1 payloads.** That is a back-compat shim. `CLAUDE.md` bans them pre-1.0 and #1840 forbids one for the wire by name: *"Do not add a serde alias or an 'accepts either shape' deserializer for the old three-key object; the plan's no-parallel-old-and-new rule applies to the wire as much as the code."* Bumping the number *without* legacy handling would be a half-signal, and adding the handling contradicts the plan. Context: moq is `OPEN` in `ARCHITECTURE.md` §Networking, nothing populates `source_processor_type` (the only `add_track` caller passes `None`), and nothing reads the version field.

**Publish the non-blank rule in the generated schema.** I agree with the goal and tried it — `#[schemars(regex(pattern = r"\S"))]` on the inner field would express `!trim().is_empty()` exactly. schemars 0.8 treats a newtype as transparent to its inner type and drops field-level validation attributes: the emitted schema is still `{"type":"string"}` with no `pattern`. I reverted rather than leave an attribute that reads as if it constrains something. Doing this properly needs hand-written `JsonSchema` *and* `ToSchema` impls; that is worth its own ticket, not a silent addition here. A `minLength: 1` would be the wrong fix — it accepts `" "`, which the validator rejects, so the published contract would disagree with the engine.

**Hold all three locks across `unregister_processor_types`' loop.** The function has zero callers — it served `remove_module`, which died with the module system (rustc reports it dead). The lock order is now documented on the struct so a future caller cannot get it wrong, but making unreachable code atomic has no beneficiary.

**"Statement has no effect" on `def process(self, ctx) -> None: ...`.** The `...` is the stub-body idiom, matching every sibling test processor in this repo (`identity_stable_processor.py` included). False positive.

### Verification

`cargo test --workspace` (excl. wheel) 1886 passed / 0 failed · wheel pytest 207 passed · engine lib 1159 · fmt and `cargo check --all-targets` clean.

PR #1855, opened 2026-08-12
docs(architecture): describe the tree that exists

tato123:
## Summary

`docs/architecture/` is governed by *"current shipped state only — no roadmap, no history of superseded designs"* (`.claude/rules/docs-policy.md`). It had drifted behind three deletions: the plugin ABI and its cdylibs (#1715), the Deno SDK and subprocess-polyglot machinery (#1715), and the schema/manifest layer (#1813 / #1854).

The result was docs instructing a reader to depend on crates, call methods and read example directories that are not in the tree.

**10 files, +205 / −254.**

## Method

One auditor per stale doc, then a **second agent per doc whose only job was to refute** every "this is gone" verdict. A false dead-verdict deletes documentation that is still correct, which is the expensive error here — so the refuter was told to default to refuting when uncertain.

**72 candidate findings → 16 refuted → 56 applied.** The 22% refutation rate is the point: those 16 would have been wrong edits. Examples of what the refuters saved — `subprocess-rhi-parity.md:174-175`, `adapter-runtime-integration.md:300-308`, `texture-registration.md:359-361`, `ray-tracing-kernel.md:58-62`, all of which describe live machinery that merely *reads* like the deleted model.

## What was wrong, by kind

| kind | example |
|---|---|
| **Crates that do not exist** | `cargo tree -p streamlib-{python,deno}-native` — an assertion the reader is told to run, naming two deleted packages. The `streamlib-adapter-<name>-helpers` convention describes a crate shape the tree has none of. |
| **Methods / constants that do not exist** | `release_for_cross_process` documented across 46 lines with Python and Deno call sites and a delegating SDK wrapper — nothing defines it. `STREAMLIB_ADAPTER_ABI_VERSION` had a section explaining its bump policy; it is nowhere. |
| **Deleted examples cited as canonical** | four `examples/polyglot-*/runner/` references; `examples/*/runner` matches nothing. |
| **Deno as a first-class target** | TypeScript samples and "both Python AND Deno together" coverage rules, in a repo where Python is the only authoring runtime. |
| **Paths that moved** | `sdk/streamlib-python/` (now the wheel), `../streamlib-consumer-rhi` (lives under `runtime/`), `escalate_request.yaml` (the wire is hand-written serde now). |
| **Dead `CLAUDE.md#` anchors** | six, four of them in the banner **every doc opens with**, pointing at an "editing markdown documentation" section CLAUDE.md has never had. |

Deletions carry the one-line marker `docs-policy.md` asks for, naming what went and why. Where a passage was merely wrong rather than dead it was corrected in place, in the doc's own voice.

## Verification

- **0 broken relative links** — every `](…)` target resolves.
- **Every `CLAUDE.md#` anchor checked against CLAUDE.md's real headings.** `#non-negotiables` is genuine and survives; the other six are fixed.
- **Every backticked repo path checked for existence.** The only remaining non-resolving ones are inside removal markers (correctly naming what was removed) and two in `vendored-vulkanalia.md` that are relative to the vendored crate root, which the surrounding sentence scopes.
- **The banner is now identical in all 8 docs that carry one.** Five never had one; none was added.

| gate | result |
|---|---|
| `xtask check-no-in-process-placement` (scans markdown) | pass — 7282 files |
| `xtask check-boundaries` | pass |
| `ship-change-removed-gate` tests | 37/37 |

No code touched — `docs/architecture/` only.

## Notes for owner

**1. Six of thirteen docs needed no content change** — `compute-kernel`, `graphics-kernel`, `ray-tracing-kernel`, `texture-readback`, `texture-ring`, `adapter-timeline-single-writer`. They describe RHI/GPU machinery the pivot never touched. This was localized rot, not a rotten directory.

**2. One adjacent section left standing, deliberately.** `adapter-authoring.md`'s `### Why not add a Vulkan device handle to the producer adapter` is rejected-alternative rationale for `release_for_cross_process` — a method that does not exist. It was outside the verified range so I did not extend the deletion into it. Two questions for you: rationale for a nonexistent API is arguably dead, and `docs-policy.md` puts "why this over the alternatives" in `docs/decisions/` rather than `docs/architecture/` anyway. Candidate for relocation or removal.

**3. The banner said "Living document. Validate, update, and critique freely."** That reads as an invitation to speculate, which is how a doc drifts from shipped state. It now states the constraint instead. If you want the invitation back, it belongs somewhere that is not the doc it licenses editing.

## Summary by CodeRabbit

* **Documentation**
  * Updated architecture guides to reflect the current shipped Python runtime and helper-process workflows.
  * Clarified subprocess communication, adapter authoring, backend selection, and GPU resource handling.
  * Removed outdated references to Deno, legacy SDKs, separate libraries, and obsolete configuration paths.
  * Standardized policy links, terminology, examples, diagrams, and current-state guidance across architecture pages.

### DIFF
--- a/docs/architecture/adapter-authoring.md
+++ b/docs/architecture/adapter-authoring.md
@@ -725,75 +725,10 @@ believe the single-pattern principle is wrong for it, **stop and
 surface the disagreement before building a parallel shape.** That
 conversation belongs in an issue, not in code.
 
-## Hypothetical walkthrough — Metal on macOS via MoltenVK
-
-Sanity check: applying the checklist to an adapter not yet shipped.
-The exercise is to confirm the checklist would produce the right
-shape mechanically.
-
-**Goal**: `streamlib-adapter-metal` exposes a host-allocated
-`VkImage` (allocated through the macOS-flavor `HostVulkanDevice`
-running on MoltenVK) as an `MTLTexture` for customers writing
-Metal-native code.
-
-Walking the checklist:
-
-1. **Crate layout** — three crates: `streamlib-adapter-metal`,
-   `streamlib-adapter-metal-helpers` (test-only), and (likely) a
-   `streamlib-adapter-metal-mtltexture-bridge` crate that holds
-   the unsafe Objective-C bridging code. Same dep-graph rules:
-   the runtime adapter crate depends on `streamlib-surface-adapter`
-   + `streamlib-consumer-rhi` + `streamlib-surface-client` +
-   `vulkanalia`, but **not** `streamlib`.
-
-2. **Module layout** — same five files (`lib.rs`, `adapter.rs`,
-   `context.rs`, `state.rs`, `view.rs`) plus a sixth `mtl.rs` for
-   the MoltenVK ↔ Metal handle conversion (analogous to
-   `-opengl/src/egl.rs`).
-
-3. **The trait impl is generic over `D: VulkanRhiDevice`.**
-   `MetalSurfaceAdapter<HostVulkanDevice>` runs in-process Rust;
-   `MetalSurfaceAdapter<ConsumerVulkanDevice>` runs cdylib-side.
-   Per-acquire is a timeline wait + a layout transition into
-   `VK_IMAGE_LAYOUT_GENERAL` plus a MoltenVK-specific call to
-   surface the underlying `id<MTLTexture>` — but the `MTLTexture`
-   handle is read-only metadata on the imported `VkImage`, not a
-   privileged op.
-
-4. **Capability marker** — a new `MetalWritable` marker exposing
-   `mtl_texture(&self) -> *const MTLTexture` (or analogous Rust-
-   side handle type). Lives in `streamlib-surface-adapter`. Existing
-   markers (`VulkanWritable`, `GlWritable`) stay untouched — the
-   adapter can also impl `VulkanWritable` if customers want to
-   issue MoltenVK Vulkan calls against the same image.
-
-5. **Tests** — conformance suite + macOS-specific round-trips.
-   Per [`.claude/rules/engine-doctrine.md`](../../.claude/rules/engine-doctrine.md),
-   cross-compile verification on Linux is required; the
-   walkthrough lands the cross-compile + native-macOS CI lane in
-   the same milestone.
-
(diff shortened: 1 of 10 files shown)


tato123:
Worked all three. Two were real errors of fact, one is a rule that contradicts itself.

**Skia in the wheel — valid, fixed.** The diagram listed `streamlib-adapter-skia` under `streamlib._engine (the wheel) ← statically linked`. `sdk/streamlib-python-wheel/Cargo.toml` does not name it, and `ARCHITECTURE.md:301` says outright that "the wheel's adapter closure excludes skia". The same diagram also declared every adapter `generic over D: VulkanRhiDevice`, which this PR had already corrected in `subprocess-rhi-parity.md` — so the diagram was carrying two false claims, not one.

**The consumer-RHI bullet — valid, and it was my regression.** My edit added "…`streamlib-adapter-opengl` imports the same host DMA-BUF through EGL rather than a consumer `VkDevice`, so it is not generic over the device flavor", and left the next bullet asserting that every subprocess imports through `ConsumerVulkanTexture` and instantiates against a consumer-flavor device. Two adjacent bullets contradicting each other, introduced by this PR. The subprocess bullet is now scoped to the Vulkan-device adapters with opengl's EGL path named.

**Roadmap / trackers / unshipped design — valid, fixed.** Docs policy bans these outright ("no tracker references, no dates, no roadmap / proposed-work"). Removed: three "will migrate in a separate issue" clauses, the `VADR-TS-002` reference, a paragraph promising a same-PR migration of two libraries, and the 70-line "Hypothetical walkthrough — Metal on macOS via MoltenVK" — whose own opening line calls it "an adapter not yet shipped", and which the plan marks post-MVP and undesigned.

**The removal markers stay — skipping four of the eight sites.** Review read them as "history of superseded designs", which the policy does ban. But the very next bullet of `.claude/rules/docs-policy.md` mandates them: *"Outright deletion is allowed when content is provably wrong — leave a one-line marker saying what was removed and why."* The rule bans history in one line and requires this history in the next. Following the explicit instruction is the defensible read, and reconciling the contradiction is a rules change, not a doc fix — it belongs in a `.claude/` PR with its own rationale, per `flow.md`.

Verification: 0 broken relative links, `xtask check-no-in-process-placement` clean over 7282 files.

